PR #3, opened 2018-06-15
Refactor reducers

joefitter:
* removed static reduers, replaced with static reducer which is hydrated from res.locals.static
* grouped data, filters and sort in datatable reducer
* removed boilerplate code in pages
* added datatable router
* removed redundant actions

TODO: only use reducers relevant to page


lib/page.js
```diff
@@ -1,56 +1,61 @@
...
+      scripts: ['/public/bundle.js', `${url}/assets/${filename}/bundle.js`],
+      Component: Component.default || Component,
+      rootReducer: combineReducers(rootReducer),
+      static: {
+        url,
+        content: merge({}, res.locals.static.content, pageContent, content)
+      }
+    });
+    res.template = path.resolve(__dirname, '../pages/common/views/index');
+    return locals(req, res, next);
```
lennym: I don't think we need this insertion point since nothing comes after it. We can just let a page mount its own middleware once initialised.

joefitter: good point 👍

lennym:
I think this is ok, but the next phase is going to be to delete all the things we don't need anymore. Do we have a rough idea what that is?

joefitter:
yep, basically all that will be left is pages, the page router, and reducers

PR #11, opened 2018-06-19
Refactor to use layout functionality in @asl/service

lennym:
* Moves index.jsx to layout.jsx so it is automatically used as the layout in apps which implement this module
* Sets the page component to res.template instead of res.locals.Component
* Removes explicit `wrap` prop from layout in favour of disabling wrapping in provider when an error is present


lennym:
The build on this will fail until it has https://github.com/UKHomeOffice/asl-service/pull/65 merged and installed as a dev-dep

joefitter:
💩

lennym:
💩💥 🔫

lennym:
👌

PR #14, opened 2018-06-20
Add highlight to changed fields in diff component

lennym:


lennym:
Added the default value `-` to empty fields here as well as it was sufficiently closely related IMO.

PR #15, opened 2018-06-21
Refactor edit journey

joefitter:


pages/common/routers/confirm.js
```diff
@@ -21,16 +15,17 @@ module.exports = ({
...
     return configure(req, res, next);
   };
 
   const _processQuery = (req, res, next) => {
     const { clear, edit } = req.query;
     if (clear) {
-      delete req.session.form[req.form.id];
-      const re = new RegExp(`/${req.form.id}/edit/confirm`);
+      delete req.session.form[req.model.id];
+      const re = new RegExp(`/${req.model.id}/edit/confirm`);
```
lennym: This needs fixing - the id, and the `edit` bit are both incorrect.

joefitter: fixed - redirects to '/'

pages/common/routers/form.js
```diff
@@ -81,14 +63,10 @@ module.exports = ({
...
-        if (some(req.form.values, (value, key) => hasChanged(value, key, defaultValues))) {
-          return next();
-        }
-        return next({ validation: { form: 'unchanged' } });
-      })
-      .catch(next);
+    if (some(req.form.values, (value, key) => hasChanged(value, key, req.model))) {
+      return next();
+    }
+    return next({ validation: { form: 'unchanged' } });
```
lennym: I think this is something we'll only want for edit forms. An unchanged empty form should throw the field level validation messages.

If this happens _after_ the field-level validation is applied then we're 👌 but I wanted to check.

PR #21, opened 2018-06-25
Remove a room

joefitter:
* Added remove router and route
* Reused confirm and success pages
* Added content


pages/place/routers/delete.js
```diff
@@ -0,0 +1,47 @@
...
+    req.session.form = req.session.form || {};
+    return next();
+  };
+
+  const _checkDelete = (req, res, next) => {
+    if (req.body.delete) {
+      req.form.delete = true;
+      return next();
+    }
+    return next(new Error('Something went wrong'));
```
lennym: Can we make the error message here actually useful please? 

They'll likely get mapped to something generic like that for _public display_ but we'll want proper errors for the logs.

pages/place/routers/delete.js
```diff
@@ -0,0 +1,47 @@
...
+    _saveValues
+  );
+
+  app.get('/',
+    _configure,
+    _checkQuery
+  );
+
+  return app;
+};
```
lennym: This feels again like we're doing a lot of things that look just like forms, and that it might be better/easier if we just make them all forms - where some have very minimal configuration/functionality rather than having lots of very similar looking routers around to make micro-optimisations of the pipeline.

pages/common/routers/form.js
```diff
@@ -18,17 +18,32 @@ const flattedNested = (data, schema) =>
...
   );
 
+const persistValues = req => {
+  merge(req.session.form[req.model.id], {
+    values: pickBy(req.form.values, (field, key) => {
+      return req.form.schema[key].persist !== false;
+    })
+  });
+};
+
```
lennym: Does anything bad happen if we persist all the values?

joefitter: nope!

PR #23, opened 2018-06-25
Use url configuration for link elements

lennym:
Instead of assuming a common url pattern across all implementations, use a url map passed to the parent app (and exposed on `res.locals.static.urls`) to go from a page key (which correspond to the folder/router structure exposed by this module) and optional parameters to build urls.

Also maps the page title content snippets to use the same structure as everything else in order to make the establishment dashboard work nicely.


lennym:
I think we should park this until after Thursday as it probably touches a few too many things for now. Let's merge it when we have some runway time to fix everything it breaks.

PR #27, opened 2018-06-26
Added comments box

joefitter:
* added field to schema on amend pages
* show on confirm page if present
* added textarea field type to form
* added 100% width to textarea

### DIFF
--- a/pages/place/update/views/confirm.jsx
+++ b/pages/place/update/views/confirm.jsx
@@ -19,9 +19,7 @@ const Confirm = ({
   errors = {},
   action = 'update',
   values,
-  model: {
-    name: modelName
-  },
+  model,
   establishment: {
     name,
     licenceNumber,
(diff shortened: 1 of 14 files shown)


pages/place/update/views/confirm.jsx
```diff
@@ -60,6 +61,14 @@ const Confirm = ({
             )
             : <Diff formatters={formatters} />
         }
+        {
+          values && values.comments && (
+            <Fragment>
+              <h2><Snippet>fields.comments.label</Snippet></h2>
+              <p>{ values && values.comments }</p>
```
lennym: This second check for `values &&` is redundant.

lennym: Do we definitely want to hide this content if no comment is provided? Or should there be some kind of "no comment provided" message. @pcuk1979?

lennym:
This mostly looks fine give-or-take my question about showing the comments on the confirm page, so will merge once we know that and the build passes.

PR #28, opened 2018-06-27
Feature - restrictions

joefitter:
* added conditional reveal component
* added restrictions when editing a room
* fixed bug with persisting data - Object.assign not merge

### DIFF
--- a/pages/common/routers/form.js
+++ b/pages/common/routers/form.js
@@ -1,6 +1,6 @@
 const { Router } = require('express');
 const bodyParser = require('body-parser');
-const { mapValues, size, some, get, isEqual, merge } = require('lodash');
+const { mapValues, size, some, get, isEqual } = require('lodash');
 const validator = require('../../../lib/validation');
 
 const defaultMiddleware = (req, res, next) => next();
@@ -12,10 +12,10 @@ const hasChanged = (value, key, item) => {
   return value !== item[key];
 };
 
-const flattedNested = (data, schema) =>
-  mapValues(schema, ({ accessor }, key) => accessor
-    ? get(data[key], accessor)
-    : data[key]
+const flattenNested = (data, schema) =>
+  mapValues(data, (value, key) => schema[key] && schema[key].accessor
+    ? get(value, schema[key].accessor)
+    : value
   );
 
 module.exports = ({
@@ -41,11 +41,10 @@ module.exports = ({
 
   const _configure = (req, res, next) => {
     req.form = req.form || {};
-    req.model = req.model || {};
-    req.model.id = req.model.id || `new-${model}`;
     req.form.schema = schema;
     req.session.form = req.session.form || {};
     req.session.form[req.model.id] = req.session.form[req.model.id] || {};
+    req.session.form[req.model.id].values = req.session.form[req.model.id].values || {};
     return configure(req, res, next);
   };
 
@@ -58,7 +57,7 @@ module.exports = ({
     if (edit) {
       return editAnswers(req, res, next);
     }
-    next();
+    return next();
   };
 
   const _clearErrors = (req, res, next) => {
@@ -67,16 +66,18 @@ module.exports = ({
   };
 
   const _getValues = (req, res, next) => {
-    req.form.values = {
-      ...flattedNested(req.model, req.form.schema),
-      ...req.session.form[req.model.id].values
-    };
+    req.form.values = Object.assign(
+      flattenNested(req.model, req.form.schema),
+      req.session.form[req.model.id].values
+    );
     return getValues(req, res, next);
   };
 
   const _process = (req, res, next) => {
-    req.form.values = mapValues(req.form.schema, ({ format }, key) => {
-      return format ? format(req.body[key]) : req.body[key] || null;
+    req.form.values = mapValues(req.body, (value, key) => {
+      return req.form.schema[key] && req.form.schema[key].format
+        ? req.form.schema[key].format(value)
+        : value || null;
     });
     return process(req, res, next);
   };
@@ -97,7 +98,7 @@ module.exports = ({
   };
 
   const _saveValues = (req, res, next) => {
-    merge(req.session.form[req.model.id], { values: req.form.values });
+    Object.assign(req.session.form[req.model.id].values, req.form.values);
     return saveValues(req, res, next);
   };
 
@@ -116,7 +117,7 @@ module.exports = ({
   const errorHandler = (err, req, res, next) => {
     if (err.validation) {
       req.session.form[req.model.id].validationErrors = err.validation;
-      merge(req.session.form[req.model.id], { values: req.form.values });
+      Object.assign(req.session.form[req.model.id].values, req.form.values);
       return res.redirect(req.originalUrl);
     }
     return next(err);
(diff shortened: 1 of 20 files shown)


pages/common/routers/form.js
```diff
@@ -67,16 +66,18 @@ module.exports = ({
...
     return getValues(req, res, next);
   };
 
   const _process = (req, res, next) => {
-    req.form.values = mapValues(req.form.schema, ({ format }, key) => {
-      return format ? format(req.body[key]) : req.body[key] || null;
+    req.form.values = mapValues(req.body, (value, key) => {
+      return req.form.schema[key] && req.form.schema[key].format
+        ? req.form.schema[key].format(value)
+        : value || null;
```
lennym: I'm not sure why the responsibility is reversed here to automatically persist _any_ keys that exist on the body. I'd keep it so it's whitelisted to only the keys that are on the schema, unless there's a compelling reason to do otherwise?

Also, that ternary is getting pretty beefy - I think I'd be happier splitting it out a bit into multiple statements.

pages/common/views/components/conditional-reveal.jsx
```diff
@@ -0,0 +1,60 @@
...
+            onChange={this.toggle}
+          />
+          <label htmlFor={`conditional-reveal-${fieldName}-no`}>{noLabel}</label>
+        </div>
+      </fieldset>
+    );
+  }
+}
+
+export default ConditionalReveal;
```
lennym: We _might_ want to roll this functionality into the radio-group component rather than maintaining the code in multiple places - i.e. give the `options` an extra property that defines any conditional content.

pages/common/views/components/inset.jsx
```diff
@@ -1,9 +1,18 @@
...
+  }, {});
+};
 
 const Inset = ({
-  children
+  children,
+  className
 }) => (
-  <div className="panel panel-border-wide">
+  <div className={classnames(concatClassNames(className))}>
```
lennym: We don't need the `concatClassNames` method at all here - `classnames` can receive an array or rest args.

This will work fine:

```
<div className={classnames('panel', 'panel-wide', className)}>
```

Alternatively sack off `classnames` altogether and just:

<div className={['panel', 'panel-wide', className].join(' ')}>

pages/common/views/containers/diff.jsx
```diff
@@ -3,7 +3,6 @@ import { connect } from 'react-redux';
 import Diff from '../components/diff';
 
 const mapStateToProps = ({ model, static: { values, schema } }) => {
-  console.log('SSS', schema);
```
lennym: 👍

pages/common/views/components/inset.jsx
lennym: Ignore that last one - it could end up with `panel panel-wide undefined`

pages/common/views/containers/diff.jsx
joefitter: 😬

pages/common/views/components/inset.jsx
joefitter: good shout

pages/common/views/components/conditional-reveal.jsx
joefitter: bucket for now I think ;)

pages/common/routers/form.js
joefitter: there _was_ a reason for this. I can't remember what it was though so I'll revert it and see if anything bad happens

pages/common/routers/form.js
```diff
@@ -41,11 +41,10 @@ module.exports = ({
 
   const _configure = (req, res, next) => {
     req.form = req.form || {};
-    req.model = req.model || {};
-    req.model.id = req.model.id || `new-${model}`;
     req.form.schema = schema;
     req.session.form = req.session.form || {};
     req.session.form[req.model.id] = req.session.form[req.model.id] || {};
+    req.session.form[req.model.id].values = req.session.form[req.model.id].values || {};
```
lennym: We might get some value out of the `set` method from lodash here.

```
set(req.session, `form.${req.model.id}.values`, {});
```

This will replace the previous three lines (I think).

Not one for _right now_, but worth considering when we're setting deep properties on things that may or may not exist.

joefitter: yeah that's a good shout - I didn't know about `set`

pages/common/views/components/form.jsx
```diff
@@ -1,34 +1,57 @@
...
+  textarea: ({ value, ...props }) => <TextArea value={value} { ...props } />,
   radioGroup: ({ ...props }) => <RadioGroup { ...props } />,
   checkboxGroup: ({ ...props }) => <RadioGroup type="checkbox" { ...props } />,
-  select: ({ ...props }) => <Select { ...props } />
+  select: ({ ...props }) => <Select { ...props } />,
+  text: props => props.value &&
+    <Fragment>
+      <h3>{ props.label }</h3>
+      <p>{ props.value }</p>
+    </Fragment>
```
lennym: We might want to consider an "arbitrary content" element, that just takes another component as a prop.

joefitter: like a `Fragment` :)

pages/common/routers/form.js
lennym: Yesterday me says "same".

Yesterday me also wonders how I'd managed to merrily use `get` for so long without wondering if `set` also existed.

pages/common/views/components/form.jsx
```diff
@@ -1,34 +1,57 @@
...
+        return conditionalReveal
+          ? <ConditionalReveal
+            fieldName={key}
+            open={model[key]}
+            label={<Snippet>{`fields.${key}.conditionalReveal.label`}</Snippet>}
+            yesLabel={<Snippet>{`fields.${key}.conditionalReveal.yesLabel`}</Snippet>}
+            noLabel={<Snippet>{`fields.${key}.conditionalReveal.noLabel`}</Snippet>}
+          >{ field }</ConditionalReveal>
+          : field;
+      })
```
lennym: I'd be way happier if this was another `if`. 10 lines seems a bit excessive for a ternary.

joefitter: I can only make one person happy a day, and today is looking like your lucky day. Please hold.

pages/common/routers/form.js
joefitter: Ah, so switching it breaks restrictions, as everything in the schema is set to req.form.values, even if it isn't included in the form - a la notes. By looping over just req.body we only set the fields sent in the form

joefitter: I guess it would be better to loop over the schema, and ignore any that aren't included in the form

pages/common/views/components/form.jsx
lennym: #hashtagblessed

pages/common/routers/form.js
lennym: I'm not sure I understand the difference between "the schema" and "the form". Isn't "the schema" the thing we use to configure the form?

joefitter: kind of, notes isn't in the form but it is in the schema, as it is in other places, like the confirm page. I have updated now to only update values received in req.body, and appear in schema

PR #34, opened 2018-06-27
fixed restrictions rendering

joefitter:


pages/common/assets/sass/_datatable.scss
```diff
@@ -103,14 +103,24 @@
...
+      ul {
+        list-style-type: disc;
+      }
+    }
+
     a {
       display: block;
       margin-left: $gutter;
       margin-bottom: 0.5em;
     }
```
lennym: TODO: make this less wildly under-specific

pages/common/assets/sass/_datatable.scss
```diff
@@ -103,14 +103,24 @@
       border-right: 1px solid #ccc;
     }
 
+    .column-two-thirds {
+      > * {
+        margin-left: $gutter-half;
+      }
```
lennym: Wouldn't it be better to pad the column rather than margin _every_ child?

pages/common/assets/sass/_datatable.scss
lennym: (by which I mean "at some point in the future" - a.k.a. not now)

pages/common/assets/sass/_datatable.scss
joefitter: yeah I did have that when it was column three quarters but that isn't a thing, so I switched to this, then changed the column type anyway. tldr: will fix, at some point

PR #36, opened 2018-06-27
Design changes to projects list

joefitter:
* removed green and orange status colours
* added snippet for 3 or more months left
* updated snippet for search box
* updated label for project title

### DIFF
--- a/pages/project/list/views/index.jsx
+++ b/pages/project/list/views/index.jsx
@@ -20,18 +20,18 @@ const ExpiryDate = ({date}) => {
   const now = moment();
   const expires = moment(date);
   const diff = expires.diff(now, 'months');
-  const className = classnames('notice', {
-    urgent: diff < 3
-  });
+  const urgent = diff < 3;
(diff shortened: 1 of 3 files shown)


pages/project/list/views/index.jsx
```diff
@@ -21,16 +21,17 @@ const ExpiryDate = ({date}) => {
...
 
   return <Fragment>
     { expires.format('DD MMMM YYYY') }
     {
       diff < 12 && (
         <span className={className}>
-          <Snippet diff={diff + 1}>{ diff === 0 ? 'diff.singular' : 'diff.plural' }</Snippet>
+          <Snippet diff={diff + 1}>{ diff > 2
+            ? 'diff.standard'
+            : diff === 0 ? 'diff.singular' : 'diff.plural' }</Snippet>
```
lennym: Can we pull the logic to determine which key we're using up to the top of the component function? 

It's a bit unwieldy to be nested inside the jsx, and makes working out what's going on with the nested ternary operators quite tricky.

PR #37, opened 2018-06-27
trim whitespace in string values

joefitter:


pages/common/routers/form.js
```diff
@@ -75,12 +75,16 @@ module.exports = ({
...
   const _process = (req, res, next) => {
     req.form.values = reduce(req.form.schema, (all, { format }, key) => {
-      if (isUndefined(req.body[key])) {
+      let value = req.body[key];
+      if (isUndefined(value)) {
         return all;
       }
+      if (typeof value === 'string') {
+        value = value.trim();
+      }
```
lennym: This will fix AS-477 (you can submit required fields with whitespace), but we can also fix AS-478 (trimming all the lines of multiline input - and so avoiding the issue we had with the whitespace in markdown) by swapping it for:

```js
value = value.split('\n').map(s => s.trim()).join('\n').trim();
```

That will be equivalent for single-line input, and will trim the whitespace from _all_ the lines of multiline input (as well as any blank lines from the start/end.

joefitter:
Your changes LGTM @lennym, merge away if you're happy

PR #42, opened 2018-07-03
Added functional tests for edit functionality

joefitter:
* added smoke tests for all pages
* added page manifest for tests
* extracted formatters for place
* fixed bug with checking changed values
* fixed bug with accessing /confirm directly
* added tests for place CRUD

### DIFF
--- a/pages/common/routers/form.js
+++ b/pages/common/routers/form.js
@@ -8,7 +8,7 @@ const defaultMiddleware = (req, res, next) => next();
 const hasChanged = ({ accessor }, newValue, oldValue) => {
   oldValue = get(oldValue, accessor, oldValue);
   if (Array.isArray(newValue)) {
-    return !isEqual([ ...newValue ].sort(), [ ...oldValue ].sort());
+    return !isEqual(newValue.sort(), oldValue.sort());
   }
   return newValue !== oldValue;
 };
--- a/pages/common/routers/form.js
+++ b/pages/common/routers/form.js
@@ -1,6 +1,6 @@
 const { Router } = require('express');
 const bodyParser = require('body-parser');
-const { mapValues, size, some, get, isEqual, reduce, isUndefined, identity, pickBy } = require('lodash');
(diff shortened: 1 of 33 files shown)


pages/common/routers/form.js
```diff
@@ -1,15 +1,16 @@
...
 
 const defaultMiddleware = (req, res, next) => next();
 
-const hasChanged = (value, key, item) => {
-  if (Array.isArray(value)) {
-    return !isEqual([ ...value ].sort(), [ ...item[key] ].sort());
+const hasChanged = ({ accessor }, newValue, oldValue) => {
+  oldValue = get(oldValue, accessor, oldValue);
+  if (Array.isArray(newValue)) {
+    return !isEqual([ ...newValue ].sort(), [ ...oldValue ].sort());
```
lennym: `.sort()` doesn't modify the array, so there's no particular reason for the `...`

joefitter: I just like the way they look

joefitter: wasn't part of this PR - but I'll update as I have the file open :)

pages/common/routers/form.js
```diff
@@ -85,19 +87,24 @@ module.exports = ({
...
     }, {});
     return process(req, res, next);
   };
 
   const _checkChanged = (req, res, next) => {
-    if (!size(req.form.schema) || some(req.form.values, (value, key) => hasChanged(value, key, req.model))) {
+    const fields = pickBy(req.form.schema, (field, key) => {
+      return field.editable !== false && Object.keys(req.model).includes(key);
+    });
+    if (!size(fields) || some(fields, (field, key) => hasChanged(field, req.form.values[key], req.model[key]))) {
```
lennym: I think this would be easier to understand if we rolled the `hasChanged` check into the `pickBy` above, so that `fields` only contains the fields in the schema _that have changed_ (and then rename `fields` to `changedFields`)

That condition is pretty wild as it is - 4 consecutive closing brackets is a good indicator of some code that needs a bit of a refactor in my eyes.

joefitter: OWZAT?

PR #45, opened 2018-07-04
Profile page amends

joefitter:
* added NACWO training section if profile is a NACWO and qualification not null
* added projects if PPL holder
* added accordion component
* fixed accordion bug when only one child


lennym:
Do we want functional tests for this?

joefitter:
do _you_ want functional tests for this?

lennym:
That depends, do you recognise a rhetorical question?

PR #47, opened 2018-07-04
Added CSRF checking

joefitter:
* added csurf lib
* added csrf middlware if env != test
* added hidden input to form and confirm pages

### DIFF
--- a/lib/middleware/csrf.js
+++ b/lib/middleware/csrf.js
@@ -1,0 +1,34 @@
+const Tokens = require('csrf');
+
+const checkSession = (req, res, next) => {
+  if (req.session && req.session.form && req.session.form[req.model.id]) {
+    return next();
+  }
+  const err = new Error('Form session not set');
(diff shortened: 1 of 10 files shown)


lib/middleware/csrf.js
```diff
@@ -0,0 +1,34 @@
+const Tokens = require('csrf');
+
+const checkSession = (req, res, next) => {
+  if (req.session && req.session.form && req.session.form[req.model.id]) {
+    return next();
+  }
+  const err = new Error('Form session not set');
+  err.code = 500;
+  return next(err);
```
lennym: You _probably_ want to set `status` rather than `code` here, but what you actually probably want is just `next(new Error('Form session not set'));` because the status defaults to 500 unless otherwise specified.

lennym:
Whoops, I didn't mean to merge this. Oh well... if you could clean up the error according to my comment above, that'd be super.

PR #52, opened 2018-07-09
bumped govuk-react-components

joefitter:


lennym:
It doesn't _really_ matter here, but I'd generally expect package.json version changes to come with a package-lock change.

lennym:
(because where the project is a top-level implementation not doing so will break the build)

joefitter:
yep, right you are, didn't add it to the commit

PR #53, opened 2018-07-09
updated content for error messages as requested

joefitter:


pages/place/content/errors.js
```diff
@@ -1,9 +1,9 @@
 module.exports = {
   site: {
-    required: 'Licensed premises all need a site name.'
+    required: 'Select a site for this new approved area.'
   },
   name: {
-    required: 'This could be a room number, like G1 for example.'
+    required: 'Enter a name for this new approved area.'
```
lennym: What if it's not new?

joefitter: You don't expect me to actually read error messages? Copy paste FTW

PR #55, opened 2018-07-09
Bugfix - nacwo showing as highlighted if not changed

joefitter:
* updated comparator to use lodash isEqual
* use accessor to compare id's for nacwo

### DIFF
--- a/pages/common/views/components/diff.jsx
+++ b/pages/common/views/components/diff.jsx
@@ -10,7 +10,7 @@ const getValue = (key, { format } = {}) => {
 const Diff = ({
   diff,
   formatters = {},
-  comparator = (...args) => !(isEqual(...args)),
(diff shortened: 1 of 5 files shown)


pages/common/views/components/diff.jsx
```diff
@@ -10,7 +10,7 @@ const getValue = (key, { format } = {}) => {
 const Diff = ({
   diff,
   formatters = {},
-  comparator = (...args) => !(isEqual(...args)),
+  comparator = (...args) => !(isEqual(args[1], args[2])),
```
lennym: `(a, b) => !isEqual(a, b)`

Not _everything_ needs a spread operator.

joefitter: FINE

PR #56, opened 2018-07-09
Bugfix - strip html tags from user input

joefitter:
* added striptags library
* added helper to parse string values in model and strip html tags

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -1,4 +1,5 @@
-const { get, isEqual } = require('lodash');
+const { get, isEqual, mapValues } = require('lodash');
+const striptags = require('striptags');
 
 const toTitleCase = str =>
   str.replace(/\w\S*/g, txt => `${txt.charAt(0).toUpperCase()}${txt.substr(1)}`);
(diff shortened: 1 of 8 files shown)


lib/utils/index.js
```diff
@@ -29,9 +30,17 @@ const hasChanged = ({ accessor }, newValue, oldValue) => {
   return newValue !== oldValue;
 };
 
+const cleanModel = model => mapValues(model, field => {
+  if (typeof field === 'string') {
+    return striptags(field);
+  }
+  return field;
```
lennym: We possibly want to traverse down through arrays and nested objects to clean _all_ strings rather than just top-level values.

pages/common/routers/form.js
```diff
@@ -98,7 +98,7 @@ module.exports = ({
       format = format || identity;
       all[key] = format(value);
       return all;
-    }, {});
+    }, {}));
```
lennym: I think I'd prefer this code if it was formatted as a number of explicit steps rather than nested function calls that are performing multiple discrete functions in a single statement. It's getting highly non-obvious.

```js

req.form.values = pick(req.body, Object.keys(req.form.schema));
req.form.values = mapValues(req.form.values, (value, key) => {
  const nullValue = req.form.schema[key].nullValue;
  return value || nullValue;
});
req.form.values = mapValues(req.form.values, (value, key) => {
  const format = req.form.schema[key].format || identity;
  return format(value);
});
req.form.values = mapValues(req.form.values, trim);
req.form.values = cleanModel(req.form.values);
```

lennym: You'd probably want to pull those out into named functions to make it even more obvious, but it took me 10 minutes of unpicking the code to work out what actual steps were being performed in there.

lennym: At the _very_ least, add the `cleanModel` as another line.

```js
req.form.values = reduce(...everything as it was before...);
req.form.values = cleanModel(req.form.values);
```

PR #59, opened 2018-07-10
Bugfix - persist conditional reveal values

joefitter:
* updated form to persist conditional reveal values
* dont validate field if conditional reveal is set to false

### DIFF
--- a/pages/common/routers/form.js
+++ b/pages/common/routers/form.js
@@ -100,6 +100,7 @@ module.exports = ({
     });
     req.form.values = mapValues(req.form.values, trim);
     req.form.values = cleanModel(req.form.values);
+    Object.assign(req.form.values, { ...pickBy(req.body, (field, key) => key.includes('conditional-reveal')) });
     return process(req, res, next);
   };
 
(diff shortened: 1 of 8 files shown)


pages/common/routers/form.js
```diff
@@ -100,6 +100,7 @@ module.exports = ({
     });
     req.form.values = mapValues(req.form.values, trim);
     req.form.values = cleanModel(req.form.values);
+    Object.assign(req.form.values, { ...pickBy(req.body, (field, key) => key.includes('conditional-reveal')) });
```
lennym: I don't think this is the right way to go about this. It's automatically whitelisting any field with `conditional-reveal`. 

I'd iterate over the schema, and do something like this to get the keys of the associated radio button.

```js
Object.keys(schema).filter(key => schema[key].conditionalReveal).map(key => `conditional-reveal-${key}`)
```

i.e. specifically find the fields which are conditional first, and then map them to their corresponding keys.

joefitter: FINE

lennym: 😢

joefitter: 🤗

PR #70, opened 2018-08-08
Error messages

pcuk1979:
Error messages done

### DIFF
--- a/pages/profile/content/index.js
+++ b/pages/profile/content/index.js
@@ -47,11 +47,11 @@ module.exports = {
       required: 'You need to enter a last name.'
     },
     email: {
-      required: 'Email address is required',
-      match: 'You need to enter an email address for this person.'
+      required: 'You need to enter an email address for this person.',
+      match: 'This email address does not exist.'
     },
     role: {
-      required: 'Select a permission level for the new user'
+      required: 'You need to set a permission level for this new user.'
     }
   },
   notifications: {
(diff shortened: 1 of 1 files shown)


pages/profile/content/index.js
```diff
@@ -41,14 +41,14 @@ module.exports = {
...
+      required: 'You need to enter a first name.'
     },
     lastName: {
-      required: 'Last name is required'
+      required: 'You need to enter a last name.'
     },
     email: {
       required: 'Email address is required',
-      match: 'Invalid email address'
+      match: 'You need to enter an email address for this person.'
```
lennym: It looks like you've edited the wrong message here. The one that's edited is for when someone has entered a value, but it's not an email address.

The "there is no value in the field" message is the line above.

pages/profile/content/index.js
```diff
@@ -41,17 +41,17 @@ module.exports = {
...
     },
     lastName: {
-      required: 'Last name is required'
+      required: 'You need to enter a last name.'
     },
     email: {
-      required: 'Email address is required',
-      match: 'Invalid email address'
+      required: 'You need to enter an email address for this person.',
+      match: 'This email address does not exist.'
```
lennym: Sorry to nitpick even harder, but what will actually be happening here is that they've entered something which isn't even an email address so I'm not sure "This email address..." makes sense.

Totally happy to merge if you say it's fine btw.

PR #72, opened 2018-08-16
Added validator to check a field matches another given field

joefitter:
Pass empty array to scripts if no scripts passed. We don't always want the common bundle

Bumped govuk-react-components

### DIFF
--- a/pages/common/assets/sass/style.scss
+++ b/pages/common/assets/sass/style.scss
@@ -74,10 +74,6 @@ th {
   display: none;
 }
 
-.grey {
-  color: $grey-1;
-}
-
 .main {
 
   .notification-summary {
--- a/pages/common/assets/sass/style.scss
+++ b/pages/common/assets/sass/style.scss
@@ -74,6 +74,10 @@ th {
   display: none;
 }
 
+.grey {
+  color: $grey-1;
+}
+
 .main {
 
   .notification-summary {
@@ -95,7 +99,7 @@ th {
       margin-bottom: $gutter;
     }
 
-    h2 {
+    h2, h3 {
       margin-bottom: $gutter/4;
       font-weight: normal;
     }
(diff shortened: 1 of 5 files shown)


pages/common/views/layouts/default.jsx
```diff
@@ -37,12 +37,14 @@ const Layout = ({
       static: { ...rest, content, urls }
     })
     : {};
-
+  if (scripts.length) {
+    scripts = ['/public/js/common/bundle.js'].concat(scripts);
+  }
```
lennym: I like this more than I was expecting to.

We just need to be careful that there's no global initialisation that ends up in the bundle.

pages/common/assets/sass/style.scss
```diff
@@ -74,6 +74,10 @@ th {
   display: none;
 }
 
+.grey {
+  color: $grey-1;
+}
+
```
lennym: Does this actually get used anywhere?

joefitter: it _did_

joefitter: In the designs there is a "Step 1 of 2" above the title on the password setup page, but as this is only one step now, I removed it

pages/common/views/components/form.jsx
```diff
@@ -9,6 +9,7 @@ import ConditionalReveal from './conditional-reveal';
 const fields = {
   inputText: props => <Input { ...props } />,
   inputEmail: props => <Input type="email" { ...props } />,
+  inputPassword: props => <Input type="password" { ...props } />,
```
lennym: I feel like there's scope here for a `type` prop and then everything is `<Input/>`. Save inventing a new top-level object every time we want a different `type` value.

This works fine too for now though.

lennym:
👍 😄

pages/common/views/components/form.jsx
joefitter: yeah, agree!

PR #79, opened 2018-08-28
Updated invite api path to /establishment/:e/invite-user

joefitter:


lennym:
Has this work been done in the API too?

joefitter:
ja

joefitter:
latest version of api has an `invite-user` endpoint

PR #81, opened 2018-08-29
Add project's expiry dates to profile of a PPL holder

lvnhmd:
hi, please review :)


joefitter:
TODO: get rid of the constants as `asl-constants` is a thing now. Out of scope for this though

PR #82, opened 2018-09-03
AS-517: Use home office frontend toolkit

wheelsandcogs:
Replaces `govuk_frontend_toolkit` with `@ukhomeoffice/frontend-toolkit` which is updated to support the new design language from `govuk-frontend`.

See:
 * https://github.com/alphagov/govuk-frontend
 * https://design-system.service.gov.uk/get-started/updating-your-code/
 * https://github.com/UKHomeOffice/frontend-toolkit


package.json
```diff
@@ -42,6 +42,7 @@
     "govuk-elements-sass": "^3.1.2",
     "govuk-react-components": "^1.2.0",
     "govuk_frontend_toolkit": "^7.4.1",
```
lennym: I'd probably expect `govuk-elements-sass` and `govuk_frontend_toolkit` to be removed from here.

lennym:
This is looking pretty good generally 👍

PR #83, opened 2018-09-04
Add landing page for PIL application

lvnhmd:


pages/pil/categories/views/index.jsx
```diff
@@ -0,0 +1,37 @@
...
+
+const Index = () => (
+  <Fragment>
+    <header>
+      <h2>&nbsp;</h2>
+      <h1>
+        <Snippet>pages.pil.categories</Snippet>
+      </h1>
+    </header>
+    <div className="grid-row">
```
wheelsandcogs: All the class names will need updating now we've merged the HO Design System work.

See: https://design-system.service.gov.uk/get-started/updating-your-code/#layout-and-grid-system-class-names

constants/index.js
```diff
@@ -16,5 +16,31 @@ module.exports = {
...
+    'E2',
+    'PPL'
+  ],
+  procedureOptions: [
+    'A. Minor/minimally invasive procedures not requiring sedation, analgesia or general anaesthesia',
+    'B. Minor/minimally invasive procedures involving sedation...',
+    'C. Surgical procedures...',
+    'D. Use of neuro...',
+    'F. Other'
+  ]
```
lennym: I think we should look to store the _data_ for these purely as `A`, `B`, `C` etc, and then store the descriptions in content.

lennym: Also, where we're adding new constants that might/will be used across different places, we probably want to keep them in https://github.com/UKHomeOffice/asl-constants

PR #89, opened 2018-09-17
Feature/edit profile

joefitter:
Update:

The profile edit screen now lives under a non-establishment specific account management page accessed via the name link in the header.

See https://projects.invisionapp.com/share/GWO35IEMKPA#/screens/320281794 for design. Currently only the "edit details" screen has been added to this page.


pages/user/update/index.js
```diff
@@ -0,0 +1,62 @@
...
+    delete req.session.form[id];
+    delete req.session.profile;
+    return res.redirect(req.originalUrl);
+  });
+
+  // app.use((req, res, next) => {
+  //   console.log(req.model.id)
+  //   console.log(req.user.id)
+  //   next();
+  // })
```
wheelsandcogs: Should we remove this?

lennym: Yes.

PR #94, opened 2018-09-20
AS-638: fix the styling and links on PIL cat selection page

wheelsandcogs:


pages/dashboard/views/index.jsx
```diff
@@ -21,6 +21,7 @@ const Index = ({
             establishments.map(est =>
               <li key={est}>
                 <Link page="establishment.dashboard" establishment={ est.id } label={ est.name } />
+                <br />
```
wheelsandcogs: Changing the styling on `ul.dashboard a` removed the `display: block` so this is just a temporary fix until we decide what the home dashboard looks like.

lennym: Can we wrap in `<p>` instead of inserting `<br/>` please? It's _mildly_ more semantic.

pages/pil/categories/views/index.jsx
```diff
@@ -6,42 +6,35 @@ import { connect } from 'react-redux';
...
-  establishment: {
-    id: estId
-  },
-  pilApplication: {
-    id: pilId
-  },
-  profile: {
-    id: profId
-  }
-}) => (
```
lennym: I wholly endorse this change! 👌

pages/dashboard/views/index.jsx
```diff
@@ -21,7 +21,9 @@ const Index = ({
             establishments.map(est =>
               <li key={est}>
                 <Link page="establishment.dashboard" establishment={ est.id } label={ est.name } />
-                <Link page="profile.invite" establishment={ est.id } label={<Snippet>pages.dashboard.invite</Snippet>}/>
+                <p>
+                  <Link page="profile.invite" establishment={ est.id } label={<Snippet>pages.dashboard.invite</Snippet>}/>
+                </p>
```
lennym: Yeah... both of them.

pages/dashboard/views/index.jsx
wheelsandcogs: done.

PR #100, opened 2018-09-24
Return to sections page after modules submit

wheelsandcogs:


pages/pil/modules/index.js
```diff
@@ -17,24 +16,6 @@ module.exports = settings => {
...
-  //     body: JSON.stringify(values)
-  //   };
-  //   return req.api(`/establishment/${req.establishment}/invite-user`, opts)
-  //     .then(() => next())
-  //     .catch(next);
-  // });
-
   app.post('/', (req, res, next) => {
     const values = req.session.form[req.model.id].values;
     console.log('TRAINING MODULES FORM VALUES ', JSON.stringify(values));
```
lennym: Can we kill this while the code is at hand?

wheelsandcogs: done.

PR #102, opened 2018-09-26
AS-640: Display training data

wheelsandcogs:


lib/utils/modules-to-certificates.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    let certificate = Object.assign({}, omit(moduleData, [
+      'id', 'module', 'migrated_id', 'species', 'not_applicable', 'other_accrediting_body'
+    ]));
+
+    certificate.hash = certificateHash;
+    certificate.modules = [{ id: moduleId, name: moduleName }];
+    certificates.push(certificate);
+  });
```
lennym: Couple of points here:

* you can still use `const` for arrays and objects that have their properties modified, or get `push`ed to. `let` vs `const` only _really_ applies to reassignment, not modification.
* a `map` (to set the hash) and a `groupBy` (to group according to hash) is probably a clearer and cleaner way of writing this code.

lennym: This _should_ roughly work:

```js
modules = modules.map(module => {
  return {
    ...module,
    hash: createCertificateHash(module)
  };
});
return groupBy(modules, 'hash');
```

PR #103, opened 2018-09-26
Allow the removal of training from a pil application

wheelsandcogs:
Ready to review.

### DIFF
--- a/pages/pil/dashboard/index.js
+++ b/pages/pil/dashboard/index.js
@@ -8,12 +8,23 @@ module.exports = settings => {
     root: __dirname
   });
 
+  const profileOwnsModules = (profile, modules) => modules.every(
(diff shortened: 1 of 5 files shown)


pages/pil/dashboard/index.js
```diff
@@ -1,12 +1,39 @@
...
 
 module.exports = settings => {
   const app = page({
     ...settings,
     root: __dirname
   });
 
+  const profileOwnsModules = (profile, modules) => modules.every(
+    moduleId => !!profile.trainingModules.find(module => module.id === moduleId)
+  );
```
lennym: Could we do something like this to reduce the number of nested functions inside functions?

```js
const profileOwnsModules = (profile, modules) => {
  const userModules = profile.trainingModules.map(m => m.id);
  return modules.every(id => userModules.includes(id));
});
```

PR #127, opened 2018-10-15
Conditional reveal

joefitter:
* bumped react-components to 0.3.0 including conditional reveal
* updated exemption schema to include child fields
* added custom validation functionality
* added custom validator to check module selected if reason empty
* split form and fieldset
* create flat model representing exemptions for ui
* post array of exemptions


pages/common/assets/sass/style.scss
```diff
@@ -295,6 +295,11 @@ dl.inline {
   }
 }
 
+.govuk-checkboxes__item > div {
+  margin-top: $govuk-gutter / 2;
+  margin-left: -24px;
+}
```
wheelsandcogs: Let's fix this in `@ukhomeoffice/frontend-toolkit`?

joefitter: is that where it goes? haha I couldn't find the css in react-components

joefitter: neither, the html is in `react-components`

joefitter: I guess that's what you meant by `ho-react-components`

joefitter: BTW, I think the css should be in the same repo as the react, `react-components` in this case

pages/pil/application/views/exemption-data.jsx
```diff
@@ -14,18 +14,10 @@ class ExemptionData extends Component {
...
-                            {module.module}
-                            {module.exemptionDescription}
-                          </li>
-                        ))
-                      }
-                    </ul>
-                  </dd>
+                  <dd>{ exemption.modules[0].module }</dd>
+
+                  <dt><Snippet>certificate.reason</Snippet></dt>
```
wheelsandcogs: This content key probably wants renaming when we do the refactor...

pages/pil/modules-exempt/index.js
```diff
@@ -1,29 +1,80 @@
...
 
   app.post('/', (req, res, next) => {
-    const values = omit(req.session.form[req.model.id].values, 'exempt');
-    values.profileId = req.profile;
-    values.modules = values.modules.map(module => ({ module, species: [] }));
-    values.exemption = true;
+    const values = req.form.values.modules.map(m => {
+      return {
+        id: req.model[`module-${m}-id`],
+        modules: [{ module: m, species: [] }],
```
wheelsandcogs: just call the mapped param `module` :D

pages/pil/modules-exempt/schema/index.js
```diff
@@ -1,10 +1,32 @@
 const { moduleCodes } = require('@asl/constants');
 const { toArray } = require('../../../../lib/utils');
+const content = require('../content');
+
+const options = moduleCodes.map(m => {
```
wheelsandcogs: `m` -> `moduleName` ?

pages/common/assets/sass/style.scss
wheelsandcogs: Yeah I was thinking that, but @lennym had a good reason to keep them separate, but now I can't remember what it was!

pages/pil/application/views/exemption-data.jsx
joefitter: yeah

PR #131, opened 2018-10-16
Success page for demo

wheelsandcogs:
This adds a redirect to the success page on submit of the PIL (doesn't actually update the PIL).

### DIFF
--- a/pages/pil/application/index.js
+++ b/pages/pil/application/index.js
@@ -4,7 +4,8 @@ const bodyParser = require('body-parser');
 module.exports = settings => {
   const app = page({
     ...settings,
-    root: __dirname
+    root: __dirname,
+    paths: ['/success']
   });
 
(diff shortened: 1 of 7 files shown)


pages/pil/application/index.js
```diff
@@ -4,14 +4,19 @@ const bodyParser = require('body-parser');
...
 
   const profileOwnsModule = ({ trainingModules, exemptions }, moduleId) => {
     return [ ...trainingModules, ...exemptions ].some(m => m.id === moduleId);
   };
 
   app.post('/', bodyParser.urlencoded({ extended: true }), (req, res, next) => {
+    if (req.body.action === 'submit-pil-application') {
+      return res.redirect(req.originalUrl + '/success');
+    }
+
```
lennym: How come we have multiple different actions all posting to the same URL?

PR #133, opened 2018-10-17
Submit the PIL application

wheelsandcogs:
This updates the PIL with a "submitted_at" time and (for now) prevents you from creating / editing a new PIL.

### DIFF
--- a/pages/pil/application/views/procedure-data.jsx
+++ b/pages/pil/application/views/procedure-data.jsx
@@ -38,13 +38,17 @@ class ProcedureData extends Component {
                 </dl>
               </div>
               <div className="govuk-grid-column-one-quarter">
-                <Link
-                  page="pil.procedures"
-                  establishment={ establishment.id }
-                  pil={ pil.id }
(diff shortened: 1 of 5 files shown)


pages/pil/application/views/procedure-data.jsx
```diff
@@ -38,13 +38,17 @@ class ProcedureData extends Component {
...
+                  <li>
+                    <Link
+                      page="pil.procedures"
+                      establishment={ establishment.id }
+                      pil={ pil.id }
+                      profile={profile.id}
+                      label={<Snippet>actions.remove</Snippet>}
+                    />
+                  </li>
+                </ul>
```
joefitter: cool list, bro

wheelsandcogs: Originally we had multiple actions per row (edit / remove) 😃

We can fix the styling to not require the lists but I don't think we should do that until we're sure the edit action is definitely not coming back for any of the sections.

PR #134, opened 2018-10-17
default profile.name to Unknown user if not set, to prevent status ba…

joefitter:
…r throwing on user lookup fail


joefitter:
@lennym poke x

PR #135, opened 2018-10-17
Allow submission of a PIL with no exemptions

wheelsandcogs:


pages/pil/exemptions/index.js
```diff
@@ -13,8 +13,10 @@ module.exports = settings => {
   app.post('/', (req, res, next) => {
 
     if (req.body.exempt === 'Yes') {
+      req.session.skipExemptions = false;
```
wheelsandcogs: Just realised this probably wants scoping to a profile id...

wheelsandcogs: ... or maybe pil id?

lennym: Profile I think.

PR #136, opened 2018-10-17
Show submit button by default

joefitter:
* updated react components to get latest reveal code
* form can now have children
* simplified application confirm component - render a form
* added validation to check declaration checked and at least one procedure selected

### DIFF
--- a/pages/common/assets/sass/style.scss
+++ b/pages/common/assets/sass/style.scss
@@ -379,3 +379,13 @@ dl.inline {
     color: $white;
   }
 }
+
+.application-confirm {
+  legend {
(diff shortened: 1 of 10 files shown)


pages/common/assets/sass/style.scss
```diff
@@ -379,3 +379,13 @@ dl.inline {
     color: $white;
   }
 }
+
+.application-confirm {
```
wheelsandcogs: We already have a selector for this (L268) so just bung the additional stuff there.

joefitter: good shout

PR #138, opened 2018-10-18
Fixed console errors, added missing keys, updated content

joefitter:


pages/pil/application/content/index.js
```diff
@@ -16,21 +16,21 @@ module.exports = {
...
   exemption: {
     skipped: 'No exemptions provided.'
   },
   procedure: {
-    category: 'Procedure category',
-    catDLabel: 'Evidence of competency',
-    catFLabel: 'Type of regulated procedure'
+    category: 'Procedure category:',
+    catDLabel: 'Evidence of competency:',
+    catFLabel: 'Type of regulated procedure:'
```
wheelsandcogs: We have to put the colons inside the snippet?

joefitter: yeah

joefitter: `<div><element>hi</element>:</div>` renders differently on the server and the client

joefitter: we might be able to get around it with a `<Fragment>` actually, so if you really don't like the above then this may work:

```js
<div>
  <Fragment>
    <Snippet>something</Snippet>
    :
  </Fragment>
</div>
```

joefitter: or perhaps wrap the `:` in a `<span>:</span>`

joefitter: One of these options is probably better

lennym:
What's the reason for the `<span>`s?

joefitter:
because server renders `your text:` and client renders `your text`. Not entirely sure why but it doesn't like a JSX element and adjacent text

joefitter:
could perhaps replace the `<span>` with a `<Fragment>` if it bugs you @lennym ?

lennym:
I was mostly just curious, _BUT_ since you mention it, a `<Fragment>` would be better if it's just a container for the sake of having a container.

PR #140, opened 2018-10-23
As 672 create a link builder component

lvnhmd:


pages/pil/application/index.js
```diff
@@ -31,8 +31,9 @@ module.exports = settings => {
       const opts = { method: 'PUT', json: { submittedAt: new Date() } };
 
       return req.api(`/establishment/${req.establishmentId}/profiles/${req.profileId}/pil/${pilId}`, opts)
-        .then(() => res.redirect(req.originalUrl + '/success'))
+        .then(() => res.redirect(req.buildRoute('pil.success', {pil: req.profile.pil.id})))
```
wheelsandcogs: When we're passing ids around, please can we append id to the params, e.g. `pilId` rather than just `pil`? This means you can easily tell if it's the object or the id.

wheelsandcogs: Actually do we just want to pass the objects into the route builder and let it grab the ids itself?

wheelsandcogs: Or maybe we support both?

pages/place/list/views/index.jsx
```diff
@@ -61,7 +61,7 @@ const Places = ({
     <FilterTable
       formatters={Object.assign({}, formatters, pageFormatters)}
       ExpandableRow={ExpandableRow}
-      createPath={allowedActions.includes('place.create') && 'place.create'}
+      createPath={allowedActions.includes('place.create') && 'place.create.create'}
```
wheelsandcogs: `place.create.create`?

wheelsandcogs: How about `place.create.new` or `place.create.start`?

pages/pil/application/index.js
lvnhmd: The name of 'pil id' needs to be pil because it is pil in the url, check out:
https://github.com/UKHomeOffice/asl-pages/blob/81986d43654bde34a7c9921ae617d79d5ae861ed/lib/middleware/route-builder.js
 
This implementation is  almost identical to the Link component https://github.com/UKHomeOffice/asl-pages/blob/master/pages/common/views/components/link.jsx

To answer your questions: 
No, as I did not chose to call it pil in the url initially.
No, as I implemented the route-builder based on the Link component.
Nope, we don't.

PR #146, opened 2018-10-30
Updated to use exemptions and certificates models rather than trainin…

joefitter:
…g-modules


pages/pil/exemptions/routers/modules.js
```diff
@@ -9,12 +9,14 @@ module.exports = () => {
   app.use('/', (req, res, next) => {
     const exemptions = req.profile.exemptions;
 
+    console.log(exemptions);
```
lennym: Remove me please.

lennym:
You've left console logs all over this.

PR #151, opened 2018-11-01
Display a tasklist

wheelsandcogs:


pages/dashboard/index.js
```diff
@@ -1,14 +1,48 @@
...
+            profileId: taskCase.data.profile.id,
+            pilId: taskCase.data.id
+          }),
+          details: taskCase.data.profile.name
+        };
+        break;
+    }
+
+    return {
+      receivedAt: moment(taskCase.updated_at).format('D MMM YYYY'),
```
joefitter: hmm, do we have this date format as a constant somewhere? I think we do, if not then we should have a central place as we use formatted dates in a few places. Project list springs to mind

pages/dashboard/views/index.jsx
```diff
@@ -1,21 +1,26 @@
...
   <Fragment>
     <header>
       <h2>&nbsp;</h2>
       <h1><Snippet name={firstName}>pages.dashboard.greeting</Snippet></h1>
     </header>
     <div className="govuk-grid-row">
-      <div className="govuk-grid-column-two-thirds">
+      <div className="govuk-grid-column-full">
+
+        <Tasklist tasks={tasks} />
```
joefitter: `tasks.length &&` <TaskList> ? to avoid rendering an empty task list if they don't have any?

joefitter: or perhaps we could handle it within the component - if there are no tasks then show a message

joefitter: It would also be worth using a connected component for TaskList rather than passing tasks, as the parent doesn't care about them and they have to be passed through...

```
// common/views/containers/task-list.js
import { connect } from 'react-redux';
import TaskList from '../components/task-list';

const mapStateToProps = ({ static: { tasks } }) => ({ tasks });

export default connect(mapStateToProps)(TaskList);
```

joefitter: this way `containers/task-list` is portable and can be used anywhere, as long as `tasks` are provided in locals

pages/dashboard/index.js
wheelsandcogs: Agreed, we should definitely have a standard format across the app. If there isn't one already I think it's a separate ticket :)

pages/common/views/components/tasklist.jsx
```diff
@@ -0,0 +1,51 @@
...
+  render() {
+    const { tasks } = this.props;
+    const outstandingCount = tasks.length;
+    const outstandingMessage = outstandingCount > 0
+      ? `You have ${outstandingCount} outstanding tasks`
+      : 'You have no outstanding tasks';
+
+    return (
+      <div className="tasklist">
+        <h2>Tasks</h2>
```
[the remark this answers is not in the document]
wheelsandcogs: I did start extracting these and then forgot about it :D

pages/dashboard/content/index.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  tasklist: {
+    title: 'Tasks'
+  }
+};
```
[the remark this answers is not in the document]
wheelsandcogs: for some reason I don't think <Snippets> in the TaskList component knew about this file - is there a way to tell it where your content is?

pages/dashboard/index.js
```diff
@@ -1,14 +1,48 @@
...
   });
 
+  const buildTask = (taskCase, req) => {
+    const licence = taskCase.data.model;
+    let action = {};
+
+    switch (licence) {
+      case 'pil':
+        action = {
+          label: 'PIL application',
```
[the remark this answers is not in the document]
wheelsandcogs: I imagine once we've got a few more cases added, they'll be something obvious we can extract to genericise it.

pages/dashboard/views/index.jsx
wheelsandcogs: The tasklist displays a message if there are no tasks, so that's already handled in the component.

The container makes sense, I'll stick it in.

pages/dashboard/index.js
wheelsandcogs: New ticket: https://jira.digital.homeoffice.gov.uk/browse/AS-769

joefitter: cool, but this line of code is new, so rather than doing it like this and saying "we'll write a ticket to fix them all" wouldn't it be better to have one less magic string? We currently define readable date formate here https://github.com/UKHomeOffice/asl-pages/blob/master/constants/index.js#L2 and we use it in other places

pages/dashboard/index.js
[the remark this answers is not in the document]
wheelsandcogs: Gonna do this when we convert it to a datatable

PR #152, opened 2018-11-05
Don't break if establishment is missing

wheelsandcogs:


lennym:
It's hard to see a reason why something would exist in this list and _not_ have an establishment, so we probably want to look at the underlying cause of this issue.

wheelsandcogs:
I believe it's because we have cases in the DB from before I started adding the establishment id to the root of the json in the `data` column, and the code is trying to use that location.

We could do with purging the existing taskflow db and seeing if any new cases have establishment id missing from there.

PR #153, opened 2018-11-05
Add multiple species

lvnhmd:


pages/common/views/components/add-another.jsx
```diff
@@ -0,0 +1,41 @@
...
+    const visible = Array(this.props.totalCount).fill(false);
+    visible.fill(true, 0, this.props.visibleCount);
+    this.setState({ visible });
+  }
+
+  addAnother(e) {
+    e.preventDefault();
+    const visible = [ ...this.state.visible ];
+    visible[visible.lastIndexOf(true) + 1] = true;
+    this.setState({ visible });
```
joefitter: you'll want to `return false` here if you don't want the `#` to end up in the URL

pages/pil/training/routers/modules.js
```diff
@@ -14,13 +14,35 @@ module.exports = settings => {
...
   app.post('/', (req, res, next) => {
     const fields = ['certificateNumber', 'accreditingBody', 'passDate', 'modules'];
     const values = pick(req.session.form[req.model.id].values, fields);
 
-    values.modules = values.modules.map(module => ({ module, species: [] }));
+    values.modules = values.modules.map(module => {
+      const species = req.form.values[`module-${module}-species`];
+      return { module,
+        species:
+         Array.isArray(species) ? species.filter(s => s !== '') : (species !== '' ? [species] : [])
```
joefitter: indentation

joefitter: also, what are you trying to do here? It looks a lot like you want `castArray` or something

pages/common/views/components/add-another.jsx
lennym: Nah, `e.preventDefault()` will have caught that.

pages/pil/training/routers/modules.js
```diff
@@ -14,13 +14,35 @@ module.exports = settings => {
...
+      };
+    }
+    );
```
joefitter: can we bring this up a line, `});`

pages/pil/training/views/modules.jsx
```diff
@@ -1,9 +1,39 @@
 import React from 'react';
 import Snippet from '../../../common/views/containers/snippet';
 import FormLayout from '../../../common/views/layouts/form';
+import Inset from '../../../common/views/components/inset';
+import AddAnother from '../../../common/views/components/add-another';
+import { Select } from '@ukhomeoffice/react-components';
+const { species } = require('@asl/constants');
```
joefitter: mixed import and require

pages/pil/training/routers/modules.js
lennym: This is some pretty wild code. Given that we always want to end up with an array, I'd do that casting first, and then filter afterwards.

i.e. `[].concat(species).filter(Boolean)` will achieve the same thing.

pages/pil/training/routers/modules.js
lennym: See my other comment 👇

joefitter:
@lvnhmd can you please not self merge unless a PR has at lease been approved?

PR #157, opened 2018-11-07
Display tasks in datatable

lvnhmd:


pages/dashboard/views/index.jsx
```diff
@@ -18,8 +30,9 @@ const Index = ({
     </header>
     <div className="govuk-grid-row">
       <div className="govuk-grid-column-full">
+        <h2><Snippet>pages.dashboard.tasks</Snippet></h2>
 
-        <Tasklist tasks={tasks} />
```
wheelsandcogs: We might wanna keep the Tasklist component as a wrapper that embeds a datatable and includes other bits, e.g. the tabs for Outstanding / In Progress / Completed and the filters for licence type?

lvnhmd: Currently there is no point of wrapping a single DataTable component into a Tasklist component, when we need to add the tabs I will do.

PR #158, opened 2018-11-09
removed components and containers in favour of @asl/components. This …

joefitter:
…has not been published yet


joefitter:
WIP - needs https://github.com/UKHomeOffice/asl-components/pull/1 merging and installing here to pass

PR #159, opened 2018-11-09
*WIP* NTCO pil endorsement

wheelsandcogs:


pages/common/views/components/pil.jsx
```diff
@@ -0,0 +1,82 @@
+import React, { Component } from 'react';
```
wheelsandcogs: This file isn't currently rendered anywhere in the current implementation but will be needed later once we render different content depending on task type. 

Do you want me to take it out, or leave it in for whoever picks up the ticket to do that work?

PR #162, opened 2018-11-19
added temporary feedback page with link to email

joefitter:
NOTE: ignored back link for now as this should be tackled for the entire app and not just this page


lennym:
👍

PR #167, opened 2018-11-21
Chore/remove sass

joefitter:
this needs https://github.com/UKHomeOffice/asl-components/pull/4 to stop WIPing


joefitter:
Also this has a bunch of new styles introduced while in progress. These will need to be moved over after this is merged

joefitter:
aaand this also needs https://github.com/UKHomeOffice/react-components/pull/13 to be merged

lennym:
I'll take the category "PRs which do not do what they say they do" for $200 please.

joefitter:
the sass is gone aint it?

lennym:
🤔

lennym:
It bothers me that we've introduced a `Panel` component and a `PanelList` component, and the latter is not multiple instances of the former.

lennym:
You have a few `TODO` comments in here. What's your intention with regard to doing those? Have you opened tickets?

joefitter:
yeah it bothers me too, I originally had it this way but getting it to work in both use cases proved to be   too big a job to include as part of this. I *really* don't like that we are using class names such as `govuk-panel__heading` when we aren't inheriting any styles too but this PR was just getting too bug to fix all the things

PR #173, opened 2018-11-23
Feature/invitation list

joefitter:


pages/profile/invitations/index.js
```diff
@@ -0,0 +1,31 @@
...
+
+  app.use((req, res, next) => {
+    const href = req.buildRoute('profile.list');
+    crumbs([
+      {
+        href,
+        label: '{{static.content.pages.profile.list}}'
+      },
+      '{{static.content.pages.profile.invitations}}'
+    ])(req, res, next);
```
lennym: Can we pull this hack for breadcrumbs in favour of a more global solution?

joefitter: yes, the reason I added it here was because it looks funky without breadcrumbs on this page - when you change the tab the page jumps down to make room for the breadcrumb

joefitter: another solution would be to pull this, _and_ the breadcrumb on the people list page

lennym: I don't really mind which, as long as this is gone. I'm in the process of putting together a scalable global solution for breadcrumbs, which this will get in the way of.

joefitter: cool I'll kill both then

PR #175, opened 2018-11-26
Add breadcrumbs to pages

lennym:
This allows each page - or set of nested pages - to push their own breadcrumb elements onto the stack.

Also removed some inline requires and pull them to the top of files.

This doesn't touch any of the PIL bread-crumbing at the moment, because it's a bit of a Balrog. There needs to be some work around how and if we expose the owner of the PIL in the UI, because it currently isn't displayed, which is probably bad.

### DIFF
--- a/pages/pil/index.js
+++ b/pages/pil/index.js
@@ -1,5 +1,10 @@
 const { Router } = require('express');
 
+const create = require('./create');
+const read = require('./read');
+const update = require('./update');
+const remove = require('./delete');
+
 module.exports = () => {
   const app = Router({ mergeParams: true });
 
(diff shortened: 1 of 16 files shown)


lennym:
This _probably_ needs https://github.com/UKHomeOffice/asl-components/pull/7 to work in any useful way.

pages/pil/index.js
```diff
@@ -16,11 +21,15 @@ module.exports = () => {
...
+  app.use((req, res, next) => {
+    req.breadcrumb('pil.base');
+    next();
+  });
 
-  app.use('/create', require('./create')());
+  app.use('/:pilId/edit', update());
+  app.use('/:pilId/delete', remove());
+  app.use('/:pilId', read());
+  app.use('/create', create());
```
joefitter: ha nice one

PR #179, opened 2018-11-29
Feature/as 774 show user permission

lvnhmd:


pages/profile/permission/content/fields.js
```diff
@@ -0,0 +1,24 @@
...
+               * Request amendments to the licences they hold.`
+      },
+      admin: {
+        label: 'Admin',
+        hint: `* View and request amendments to all licences at the establishment.
+               * Manage user accounts and permissions at the establishment.`
+      }
+    }
+  }
+};
```
lennym: This content already exists for inviting a user. We should only have one copy of it, and re-use accordingly.

pages/profile/permission/schema/index.js
```diff
@@ -0,0 +1,14 @@
+const PERMISSION_LEVELS = ['admin', 'read', 'basic'];
```
joefitter: I _think_ we have these defined as constants in @asl/constants

pages/profile/permission/content/index.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  fields: require('./fields'),
+  buttons: {
+    submit: 'Save',
+    remove: 'Remove'
+  },
+  title: 'Change permission level',
+  remove: 'Remove permissions for this establishment',
+  copy: 'If you remove the user from this establishment they will no longer be able to view any of this establishment\'s data and all currently held licenses will need to either be transferred to another establishment or put on hold'
```
lennym: I'm not sure this content corresponds with the technical reality. It's probably more a question for @pcuk1979 but I think we need them to remove/revoke associated roles/licences _before_ being able to remove the user.

pages/profile/permission/views/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+
+const Page = ({ url }) => {
+  return (
+    <Fragment>
+      <FormLayout formatters={formatters}>
+        <header>
+          <h1>
+            <Snippet>title</Snippet>
+          </h1>
+        </header>
```
joefitter: we have a <Header /> component in @asl/components to inherit shared styles

pages/profile/read/views/index.jsx
```diff
@@ -41,6 +41,8 @@ const getPremises = roles => {
 
 const Index = ({
   isUser,
+  isAdmin,
+  profileRole,
```
lennym: This shouldn't be checking `isAdmin` but that the user has permissions to perform the `profile.permissions` action.

`user.allowedActions.includes('profile.permissions')` (or similar) should do it.

Take a look at the establishment dashboard for an example.

lennym: Here's an example of using `allowedActions` https://github.com/UKHomeOffice/asl-pages/blob/master/pages/place/list/views/index.jsx#L54-L72

pages/profile/permission/views/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+      </FormLayout>
+      <Fragment>
+        <h2>
+          <Snippet>remove</Snippet>
+        </h2>
+        <p>
+          <Snippet>copy</Snippet>
+        </p>
+        {/* <ApplyChanges type="form" method="POST" action={`${url}/remove`}> */}
```
lennym: If this isn't needed then remove it please.

lennym: Or is this what you meant by 

> I have temporarily disabled remove button

?

lvnhmd:
Yes

pages/profile/permission/content/fields.js
lvnhmd: @lennym there are slight copy differences, I will ignore them

pages/profile/permission/content/index.js
lvnhmd: To confirm : do we need to amend the copy ?

pages/profile/permission/views/index.jsx
lvnhmd: Yes.

pages/profile/permission/views/index.jsx
lvnhmd: Replaced.

pages/profile/permission/content/fields.js
lennym: 👍 At this point I think it's fair to assume those are accidental and assume consistency. We can change later if we need to.

pages/profile/permission/content/index.js
lvnhmd: This is ticket https://jira.digital.homeoffice.gov.uk/browse/AS-834.

pages/profile/permission/index.js
```diff
@@ -0,0 +1,56 @@
...
+module.exports = settings => {
+  const app = page({
+    root: __dirname,
+    ...settings
+  });
+
+  app.use('/', form({ schema }));
+
+  app.use('/', (req, res, next) => {
+    res.locals.static.isNamed = !!((req.profile.roles && req.profile.roles.length) || req.profile.pil || (req.profile.projects && req.profile.projects.length));
```
lennym: I feel like this condition probably wants to be broken down into a couple of lines rather than trying to do the whole thing in one.

Also, we'll want to check that the licences held are active as well as just existing.

Also also, I'm not 100% sure of the scope of the profile at this point but we'll likely need to check if the profile's PIL and projects are associated with the current establishment. So if a user has projects or a PIL at establishment X, then they should still be removable from establishment Y. Is this the case?

pages/profile/permission/content/index.js
```diff
@@ -0,0 +1,18 @@
...
+      label: ''
+    }
+  },
+  buttons: {
+    submit: 'Save',
+    remove: 'Remove'
+  },
+  title: 'Change permission level',
+  remove: 'Remove permissions for this establishment',
+  warning: 'If you remove the user from this establishment they will no longer be able to view any of this establishment\'s data and all currently held licenses will need to either be transferred to another establishment or put on hold',
```
lennym: The content about "all currently held licences" is redundant now on the basis that if they have currently held licences then this copy is not shown.

lvnhmd: You mean the warning ? But if they don't it is shown, i.e. it is shown for Read Only user who can 
be removed.

lennym: I mean we should remove all the content after "be able to view any of this establishment\'s data" because a) we already know they don't have any current licences, and b) it's referring to functionality that doesn't exist and likely will never exist - i.e. putting licences "on hold".

pages/profile/permission/index.js
lvnhmd: I just made sure that profile's roles, PIL and projects are scoped under establishment.Check 
https://github.com/UKHomeOffice/asl-schema/pull/119.

pages/profile/permission/index.js
```diff
@@ -0,0 +1,59 @@
...
+    ...settings
+  });
+
+  app.use('/', form({ schema }));
+
+  app.use('/', (req, res, next) => {
+    res.locals.static.isNamed =
+      !!((req.profile.roles && req.profile.roles.length) ||
+         (req.profile.pil && req.profile.pil.status === 'active') ||
+         (req.profile.projects && req.profile.projects.length));
```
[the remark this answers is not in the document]
lennym: Also, could we maybe split the conditions into separate statements? 

i.e.

```js
const hasRoles = ...
const hasPil = ...
const hasProjects = ...
res.locals.static.isNamed = hasRoles || hasPil || hasProjects;
```

pages/profile/permission/index.js
```diff
@@ -0,0 +1,59 @@
...
+    const values = {
+      role: req.session.form[req.model.id].values.role,
+      establishmentId: req.establishmentId,
+      profileId: req.profileId
+    };
+
+    const opts = {
+      method: 'PUT',
+      headers: { 'Content-type': 'application/json' },
+      body: JSON.stringify(values)
```
lennym: FYI, you can do `json: values` here and not need the content type header or the `JSON.stringify`.

lennym:
It _looks like_ we're only hiding the link for users who don't have the correct permissions, and not preventing access or submission to the actual page where they adjust permissions. Is that right?

lvnhmd:
> It _looks like_ we're only hiding the link for users who don't have the correct permissions, and not preventing access or submission to the actual page where they adjust permissions. Is that right?

Nope, we are preventing access to the actual page: https://github.com/UKHomeOffice/asl-pages/pull/179/commits/40f7f10375c7e7a154136cca5d26b7b90e780b47

lennym:
I see, my bad.

We _might_ want to add something that prevents access to that page for a person's own profile, especially if it's not also blocked at an API submission level.

pages/profile/index.js
```diff
@@ -42,6 +43,19 @@ module.exports = () => {
     next();
   });
 
+  app.use('/:profileId/permission', permissions('profile.permissions'),
+    (req, res, next) => {
+      Promise.resolve().then(() => {
+        if (req.user.profile.id === req.profileId) {
+          const err = new Error('Unauthorised');
+          err.status = 401;
```
lennym: _Technically_ this should be a 403 rather than a 401, but it doesn't _really_ matter.

pages/profile/index.js
```diff
@@ -42,6 +43,19 @@ module.exports = () => {
...
+          next(err);
+        }
+      });
+      next();
+    },
```
lennym: I'd move this code into the permission router as top-level middleware. That way all the logic for that page is encapsulated there.

pages/profile/index.js
lvnhmd: Wait a min , why ? 401 Unauthorized vs 403 Forbidden ?

lennym: 401 is for when a person can't do a thing because we don't know enough about them or there is an action they need to perform first, i.e. they're not logged in.

403 means "we know exactly who you are, and you are not allowed to do this".

In this case we know and recognise the user, it's just that they're not allowed to perform the action in question, and there isn't anything they can do to be allowed to.

lvnhmd: Ah, I see, thank you.

PR #180, opened 2018-11-29
updated service, use store from service in page.js

joefitter:


lennym:
I like it.

lennym:
Although, what is `res.locals.rootReducer` actually used for?

And if it's always the same, then why not just import it in the template?

joefitter:
seems legit

pages/common/views/base.jsx
```diff
@@ -3,6 +3,7 @@ import omit from 'lodash/omit';
 import { createStore } from 'redux';
 import { Provider } from 'react-redux';
 import HomeOffice from '@asl/service/ui/components/home-office';
+import rootReducer from '@asl/service/ui/reducers';
```
lennym: Next question... if all of the things that this template needs to render exist in `@asl/service/ui`, why is this template not also there?

a.k.a. Make `@asl/pages` _only_ be pages.

joefitter: I was trying to stay in my lane

joefitter: you'd be a terrible driving instructor

lennym: Or rather, by default `@asl/service` uses https://github.com/UKHomeOffice/asl-service/blob/master/ui/views/layout.jsx as the layout, which we override to use this template, so we can add all the custom stores and reducers.

Except now those aren't _custom_ stores and reducers, they're just stores and reducers, so why not make that file this file?

lennym: Also, I'm happy to merge this, and then do what I describe later, but it seems a logical next step.

What _is_ your lane btw?

joefitter: my lane was the console errors and broken-ness on the dashboard

joefitter: and yeah, I'm totally for pulling everything that isn't a page out of pages, especially page.js

lennym: Do you want me to merge this then, or will you crack on? Or both?

joefitter: I'd rather get this in, as it should fix the dashboard issue, then I will do the things

joefitter: as the things are likely to break the other things

PR #181, opened 2018-11-29
Fix styling of form control in remove place journey

lennym:
Use the same components as the place amendment journey for removal to ensure the correct markup is applied.


joefitter:
As a note - we should probably just use a form and form router for this - we'd get the csrf stuff et al for free then

joefitter:
I'll add a ticket

PR #190, opened 2018-12-03
if sort and accessor defined, only use sort for sorting

joefitter:


pages/project/list/schema/index.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
   },
   licenceHolder: {
     show: true,
+    sort: 'licenceHolder.lastName',
     accessor: 'licenceHolder.name',
```
lennym: We might want to get rid of this anyway, and use a formatter to render the full name from the model instead on account of the fact that the `name` virtual property is flaky, and should be considered deprecated.

joefitter: yes agree, however the `sort` property is needed as we need to sort on lastName

joefitter: that was my way of saying "I'm staying in my lane", but actually I'm gunna swerve into oncoming traffic and fix this. Please hold.

lennym: Shall I merge then?

pages/project/list/views/index.jsx
```diff
@@ -12,7 +12,7 @@ import {
 
 export const formatters = {
   licenceHolder: {
-    format: (name, project) => <Link page="profile.view" profile={project.licenceHolder.id} label={ name } />
+    format: ({ id, firstName, lastName }) => <Link page="profile.view" profile={id} label={`${firstName} ${lastName}`} />
```
lennym: 👍

PR #194, opened 2018-12-04
removed page.js - import from asl/service, removed reducers and store

joefitter:


lennym:
There would have been a case for leaving `./lib/page.js` in place and aliasing to `@asl/service` to make this a less break-y change, but it's done now.

You'll just need to be careful when updating the UI services.

joefitter:
yeah I thought of that, but then I also thought it would be "cleaner" to move it all together and make sure it's bumped, otherwise we get left with things that aren't used anymore

PR #207, opened 2018-12-12
Display the correct content when an application is rejected

wheelsandcogs:


pages/task/read/content/confirm.js
```diff
@@ -4,17 +4,25 @@ module.exports = {
...
       declaration: {
-        title: 'Declaration',
-        text: `By endorsing this application, I agree that:
-           * I have the authority of the establishment licence holder.
-        `
+        title: 'Declaration'
       },
       link: {
-        exit: 'Exit'
+        exit: 'Save and exit'
```
joefitter: I may be wrong, but I think this link text is used (or should be used) on all forms, so it ideally should be pulled up a few levels to the base content

lib/utils/index.js
```diff
@@ -76,6 +76,14 @@ const buildModel = (...args) => {
 
 };
 
+const licenceCanProgress = decision => {
+  return [
+    'ntco-endorsed',
+    'inspector-recommended',
+    'resolved'
+  ].includes(decision);
+};
```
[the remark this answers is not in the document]
wheelsandcogs: It's true for those specific states, and they can only get to those states depending upon which licence type they're reviewing, so I think we should be ok. There may end up being more states in this list but we don't know what they are atm.

pages/task/read/content/confirm.js
wheelsandcogs: Makes sense, but sounds like a lane change - let's create a ticket.

pages/task/read/views/confirm.jsx
```diff
@@ -13,13 +14,16 @@ const Confirm = ({ task }) => {
           <span className="action">
             <Link page="task.read" taskId={task.id} label={<Snippet>task.confirm.decision.change</Snippet>} />
           </span>
-          <p>Endorsed</p>
+          <p><Snippet>{`task.${decision}.decision`}</Snippet></p>
+          { reason && <p>{reason}</p> }
```
joefitter: shouldn't we be using a snippet here?

wheelsandcogs: For the reason? It's user-entered text.

lib/utils/index.js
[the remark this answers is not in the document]
joefitter: 👍

pages/task/read/views/confirm.jsx
joefitter: cool beans

pages/task/read/content/confirm.js
lennym: It is used on a lot of forms, but also it's fictional functionality so we should avoid it IMO.

joefitter: good point...

lennym: We might as well label the button "click this and we'll lose your data".

pages/task/read/routers/confirm.js
```diff
@@ -11,7 +11,12 @@ module.exports = () => {
 
   app.use(form({
     locals: (req, res, next) => {
+      const formValues = req.session.form[`${req.task.id}-decision`].values;
+      const decision = formValues.decision;
+
       res.locals.static.task = req.task;
+      res.locals.static.decision = decision;
+      res.locals.static.reason = formValues[`${decision}-reason`];
```
lennym: Does this need to be sent down in the locals or can it be deduced on the client?

I'd like to try to avoid dumping too many things onto locals if they can be calculated from existing values. The script logic is cacheable, but adding locals is not cached because they're sent inline in the page.

PR #209, opened 2018-12-12
Feature/as 800 display task content

lvnhmd:


pages/task/read/views/pil.jsx
```diff
@@ -0,0 +1,73 @@
+import React, { Fragment } from 'react';
```
wheelsandcogs: We might want to move the PIL component up to common because it will probably be useful on other pages?

pages/task/read/content/index.js
```diff
@@ -9,11 +9,21 @@ module.exports = {
...
+      label: 'Do you endorse this application?'
     },
     reason: {
       label: 'Reason for rejection'
-    }
+    },
+    options: [
+      {
+        label: 'Yes',
+        hint: 'I confirm that the applicant holds the neccessary training or experience to carry out the categories of procedures listed in this applciation',
```
lennym: Typo here: `application`

pages/task/read/index.js
```diff
@@ -17,6 +19,17 @@ module.exports = settings => {
...
+  app.use('/', (req, res, next) => {
+    if (req.task.data.data.profileId && req.task.data.data.establishmentId) {
+      return req.api(`/establishment/${req.task.data.data.establishmentId}/profile/${req.task.data.data.profileId}`)
+        .then(({ json: { data } }) => {
+          req.profile = cleanModel(data);
+        })
+        .then(() => next())
+        .catch(next);
+    }
+  });
```
joefitter: what happens if `req.task.data.data.profileId && req.task.data.data.establishmentId` is falsy?

constants/index.js
```diff
@@ -11,5 +11,15 @@ module.exports = {
         prolonged general anaesthesia.`,
     D: `Use of neromuscular blocking agents`,
     F: 'Other'
+  },
+  pilEndorsmentStatuses: {
+    'ntco-endorsed': {
+      label: 'Yes',
+      hint: `I confirm that the applicant holds the neccessary training or experience to carry out the categories of procedures listed in this applciation`
```
lennym: We probably shouldn't use backticks for content which isn't template literals.

pages/task/read/index.js
joefitter: hint: nothing, it will spin _forever_

pages/task/read/views/pil.jsx
lennym: Leave it where it is for now. If and when it becomes useful elsewhere it can be moved.

pages/task/read/index.js
```diff
@@ -66,11 +79,13 @@ module.exports = settings => {
         }, {})
       };
 
+      req.form.schema.decision.options = merge(req.form.schema.decision.options, res.locals.static.content.fields.options);
```
joefitter: This seems odd, you are mutating options and assigning the value to options, I would do one or the other

pages/task/read/views/index.jsx
```diff
@@ -70,23 +71,25 @@ const Task = ({ task }) => {
...
+            <li><a href="#endorse">Endorse Application</a></li>
           </ol>
         </div>
 
         <div className="govuk-grid-column-two-thirds">
-          <Form formatters={formatters} />
+          <Pil />
+          <div className="govuk-grid-column-two-thirds" id="endorse">
+            <Form formatters={formatters} />
+          </div>
```
lennym: The nesting of these grid elements seems a bit off. We're nesting a lot of `two-thirds` containers inside each other, which we almost certainly don't want.

pages/task/read/index.js
joefitter: also it feels a little icky that we are directly mutating the schema here - we should find a better way, ideally lookup content in the component

pages/task/read/views/pil.jsx
```diff
@@ -0,0 +1,73 @@
...
+import moment from 'moment';
+import { Snippet } from '@asl/components';
+import { dateFormat, procedureDefinitions } from '../../../../constants';
+import { connect } from 'react-redux';
+
+const Pil = ({ profile }) => {
+  const pil = profile.pil;
+  return (
+    <div className="govuk-grid-row">
+      <div className="govuk-grid-column-three-quarters" id="training">
```
lennym: I don't think we should have the grid containers in here. The component should _only_ define itself and not the context in which it exists.

constants/index.js
lennym: Same typo here too: `application`

pages/task/read/content/index.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   task: {
-    title: 'New task',
+    title: 'Review PIL Application',
```
wheelsandcogs: Just thinking, this content is all PIL specific but will need to change depending upon what type of task is being actioned.

When I created the ticket I was thinking we would put the structure in place to do that now, but I'm open to leaving it until we implement another task other than PIL endorsement?

constants/index.js
lennym: Actually, I'm not sure why this content is here at all. Given it's content, and is also defined in the content files. Remove it.

pages/task/read/content/index.js
lennym: I'd have the logic exist for it to be able to change, but leave the other branches empty.

That is, have a _way_ to define dynamic content based on task type/state, but don't actually define any others.

lennym: I would say yes.

pages/task/read/views/pil.jsx
```diff
@@ -0,0 +1,71 @@
...
+      </div>
+
+      <div id="procedures">
+        <h2><Snippet>pil.procedures.title</Snippet></h2>
+        <h3><Snippet>pil.procedures.categories</Snippet></h3>
+        { pil.procedures.length && (
+          <Fragment>
+            { pil.procedures.map((procedure, index) => (
+              <Fragment key={index}>{procedure.toUpperCase()}.&nbsp;{procedureDefinitions[procedure]}<br />
+                <br /></Fragment>
```
joefitter: this HTML is a little but wild. Firstly we are wrapping a string in a fragment, which will render just a plain old tagless string - this is probably bad. Also the use of `&nbsp;` is a little strange - you could just us a space. Last moan would be the use of constants to store content strings - these should be content. Although I do realise the latter was introduced before this PR

joefitter: or even better - a snippet with passed in params :)

pages/task/read/index.js
```diff
@@ -17,10 +18,26 @@ module.exports = settings => {
...
+    next();
+  });
+
   app.use('/', form(Object.assign({
     configure: (req, res, next) => {
       req.schema = schemaGenerator(req.task);
 
+      const key = req.task.data.model;
+      // move the content[key] properties one up
+      Object.assign(res.locals.static.content, res.locals.static.content[key]);
```
joefitter: what's the intention here? The added comment doesn't describe the code - it isn't being 'moved' it is being copied, and it seems like data manipulation for the sake of data manipulation. This should be a concern of the template - I assume the 'key' or 'model' property is exposed in the template so you could just use it in a Snippet:

```
<Snippet>{`${key}.property`}</Snippet>
```

pages/task/read/views/pil.jsx
```diff
@@ -0,0 +1,70 @@
...
+    <Fragment>
+      <h2><Snippet>pil.training.title</Snippet></h2>
+      { profile.certificates.length && profile.certificates.map((certificate, index) => (
+        <div className="certificate" key={index}>
+          <Fragment>
+            <h3><Snippet>pil.training.certificate.details</Snippet></h3>
+            <Snippet>pil.training.certificate.number</Snippet><span>:</span> {certificate.certificateNumber}<br/>
+            <Snippet>pil.training.certificate.awarded</Snippet><span>:</span> {moment(certificate.passDate, 'YYYY-MM-DD').format(dateFormat.short)}<br/>
+            <Snippet>pil.training.certificate.expiry</Snippet><span>:</span> {moment(certificate.passDate, 'YYYY-MM-DD').add(5, 'years').format(dateFormat.short)}<br/>
+            <Snippet>pil.training.certificate.body</Snippet><span>:</span> {certificate.accreditingBody}<br/>
```
joefitter: Why are we hard-coding a date format, _and_ using a constant?

joefitter: also, we have the lib `date-fns` for use in templates - moment is mahoosive so we removed it from anything that gets bundled

pages/task/read/views/pil.jsx
```diff
@@ -0,0 +1,70 @@
...
+import { dateFormat, procedureDefinitions } from '../../../../constants';
+import { connect } from 'react-redux';
+
+const Pil = ({ profile }) => {
+  const pil = profile.pil;
+  return (
+    <Fragment>
+      <h2><Snippet>pil.training.title</Snippet></h2>
+      { profile.certificates.length && profile.certificates.map((certificate, index) => (
+        <div className="certificate" key={index}>
```
joefitter: we are using a className here and an id below for 'exemptions' and 'procedures'. My question though is do we need either? Are they actually used for anything?

joefitter: I ask as I can't see any styles being added or these being referenced anywhere so seems a bit like clutter.

pages/task/schema/index.js
```diff
@@ -17,10 +19,14 @@ module.exports = (task) => {
...
+  const key = task.data.model;
+
   const options = task.nextSteps.map(nextStep => {
+    const option = content[key].fields.options.find(opt => opt.value === nextStep.id);
     return {
-      value: nextStep.id,
-      label: nextStep.id,
+      value: option.value,
+      label: option.label,
+      hint: option.hint,
```
joefitter: If we are using content like this, then we are doing something wrong. Radio groups and checkbox groups should lookup option labels and hints the same way as inputs lookup labels.

pages/task/read/index.js
lvnhmd: Copy the _content[key]_ properties one up as they are currently in **pil{}** and we want them in **content{}**, as they will be looked up at content level in _asl-components(fieldset)_. The idea is when we have a task of another type - not just pil application - we can just drop in a file with the task's content - e.g. _content/ppl.js_ and require it in _content/index.js_ .

joefitter: why don't we... drumroll... change the content?

pages/task/read/views/pil.jsx
lennym: It _might_ be that it's worth adding some references to assist in integration testing, but 100% agree that if that's what we're doing then they should be consistent (and possibly use a different attribute other than `className`- i.e. `data-*` -  to disambiguate from classes applied to styling)

pages/task/read/index.js
lennym: I agree with @joefitter - let's not do this, and deal with the dynamic lookups on the client.

The only place I can see where that's actually needed is for the page title anyway.

pages/task/read/views/index.jsx
```diff
@@ -70,23 +71,24 @@ const Task = ({ task }) => {
...
+            <li><a href="#endorse">Endorse Application</a></li>
           </ol>
         </div>
 
         <div className="govuk-grid-column-two-thirds">
-          <Form formatters={formatters} />
+          <Pil />
+          <div id="endorse">
+            <Form formatters={formatters} />
+          </div>
```
lennym: This looks a bit odd, that we have PIL specific stuff in the left column, and then a `<PIL />` component in the right column.

It would probably be better to have the `<PIL />` component deal with all the PIL things across two columns, and then a separate row/columns underneath for the common form part.

lennym: I'm happy for this to be left until the next implementation though, and not resolved in this PR.

pages/task/read/index.js
joefitter: probably, I admit I haven't actually looked at how this is used, so I'm whinging about something I don't know anything about

lvnhmd: No, it is also needed for the fields labels and hints in [fieldset](
        asl-components/src/fieldset/index.jsx
      ).I think Joe means switch by type ?

lvnhmd: No, it is also needed for the fields labels and hints in [fieldset](
       https://github.com/UKHomeOffice/asl-components/blob/master/src/fieldset/index.jsx
      ).I think Joe means switch by type ?

lennym: 👍 

I nevertheless maintain that it's probably better to define a prefix to do the lookup using the full path rather than copying _all_ of the content around. Especially since _right now_ the only scope is "PIL or not-PIL" but there will also be dependencies on the task type and status that eventually need to be considered.

pages/task/read/views/pil.jsx
lvnhmd: That is right, we do not need the classNames, I removed them. We need the ids to be able to [navigate](https://github.com/UKHomeOffice/asl-pages/blob/439b46d5f065aa2be090059c85f8db80c4a68d1a/pages/task/read/views/index.jsx#L72).

pages/task/read/views/index.jsx
wheelsandcogs: We'll probably want a `SectionNavigation` component at some point that you can pass section props to (map of labels & section ids?) that can trigger a scroll to the section - it's gonna be needed for the NTS review page, and probably other bits as well.

pages/task/read/content/index.js
```diff
@@ -1,27 +1,12 @@
...
-      customValidate: 'Please provide a reason'
-    }
+const base = require('./base');
+const { merge } = require('lodash');
+
+module.exports = (task) => {
+
+  switch (task.data.model) {
+    case 'pil':
+      return merge({}, base, require('./pil'));
```
lennym: Can you move the require up to the top of the file please?

pages/task/read/index.js
```diff
@@ -17,8 +20,21 @@ module.exports = settings => {
...
+        })
+        .then(() => next())
+        .catch(next);
+    }
+    next();
+  });
+
   app.use('/', form(Object.assign({
     configure: (req, res, next) => {
+      res.locals.static.content = merge({}, getContent(req.task), res.locals.static.content);
```
lennym: You _probably_ want to swap the order of the last two arguments. The most specific content should be last to be merged on.

pages/task/read/views/pil.jsx
```diff
@@ -0,0 +1,72 @@
...
+        <h2><Snippet>pil.procedures.title</Snippet></h2>
+        <h3><Snippet>pil.procedures.categories</Snippet></h3>
+        { pil.procedures.length && (
+          <Fragment>
+            { pil.procedures.map((procedure, index) => (
+              <p key={index}>{`${procedure.toUpperCase()}. ${procedureDefinitions[procedure]}`}</p>
+            ))}
+          </Fragment>
+        ) }<br />
+        <hr /><br />
```
lennym: I don't think there's ever a good reason to have `<br/>`s hardcoded in templates. Can we remove them, and then we can clean the styles to achieve the same effect later.

pages/task/read/views/pil.jsx
```diff
@@ -0,0 +1,72 @@
...
+      <h2><Snippet>pil.training.title</Snippet></h2>
+      { profile.certificates.length && profile.certificates.map((certificate, index) => (
+        <div key={index}>
+          <Fragment>
+            <h3><Snippet>pil.training.certificate.details</Snippet></h3>
+            <Snippet>pil.training.certificate.number</Snippet><span>:</span> {certificate.certificateNumber}<br/>
+            <Snippet>pil.training.certificate.awarded</Snippet><span>:</span> {formatDate(certificate.passDate)}<br/>
+            <Snippet>pil.training.certificate.expiry</Snippet><span>:</span> {formatDate(addYears(certificate.passDate, 5))}<br/>
+            <Snippet>pil.training.certificate.body</Snippet><span>:</span> {certificate.accreditingBody}<br/>
+            <Snippet>pil.training.certificate.file</Snippet><span>:</span> <br/>
```
lennym: Can we use a `<dl>` instead of random line breaks please?

At the absolute worst use `<p>` but a `<dl>` is best.

constants/index.js
```diff
@@ -1,7 +1,8 @@
 module.exports = {
   dateFormat: {
     short: 'D/M/Y',
-    medium: 'DD MMMM YYYY'
+    medium: 'DD MMMM YYYY',
+    ddMMyyyy: 'DD/MM/YYYY'
```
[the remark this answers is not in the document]
lennym: > short, medium and ddMMyyyy

The three genders.

In all seriousness, if we have a number of defined date formats we should stick to using them consistently and not just keep adding new formats every time one appears in a design.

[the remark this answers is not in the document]
lennym: You've already done that though, so 👍 👏

PR #210, opened 2018-12-12
Allow a default row limit to be set for datatables

wheelsandcogs:


pages/place/list/index.js
```diff
@@ -9,13 +9,15 @@ module.exports = settings => {
...
+  const defaultRowCount = 50;
+
   app.use(datatable({
     filters: Object.keys(pickBy(schema, s => s.filter)),
     getApiPath: (req, res, next) => {
       req.datatable.apiPath = `/establishment/${req.establishmentId}/places`;
       next();
     }
-  })({ schema }));
+  })({ schema, defaultRowCount }));
```
lennym: I'm not sure there's any need for the `const` to be defined and only used once. I'd just define inline.

wheelsandcogs: It's just so it's obvious what the number is rather than it just being a nameless int, but I can remove.

wheelsandcogs: heh, talking nonsense, fixed.

PR #214, opened 2018-12-20
Allow a datatable schema to be provided as part of the configure

wheelsandcogs:
This allows dynamic replacement of the schema for a datatable, see: 
https://github.com/UKHomeOffice/asl-internal-ui/pull/40/files#diff-8fcd7a31470eb38e82be8f8cc7f5a094R30

### DIFF
--- a/pages/common/routers/datatable.js
+++ b/pages/common/routers/datatable.js
@@ -94,6 +94,10 @@ module.exports = ({
   };
 
   const _getValues = (req, res, next) => {
+    if (req.datatable.disable) {
+      return next();
+    }
+
     const query = buildQuery(req, req.datatable.schema);
     const apiPath = castArray(req.datatable.apiPath);
 
--- a/pages/common/routers/datatable.js
+++ b/pages/common/routers/datatable.js
@@ -41,16 +41,19 @@ module.exports = ({
   apiPath,
   schema,
   defaultRowCount = 10
-}) => {
+} = {}) => {
   const app = Router();
 
   const _configure = (req, res, next) => {
     req.datatable = req.datatable || {};
+    req.datatable.schema = schema;
     return configure(req, res, next);
   };
 
   const _getApiPath = (req, res, next) => {
-    req.datatable.apiPath = apiPath;
+    if (apiPath) {
+      req.datatable.apiPath = apiPath;
+    }
     return getApiPath(req, res, next);
   };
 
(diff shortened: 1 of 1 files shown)


pages/common/routers/datatable.js
```diff
@@ -41,16 +41,19 @@ module.exports = ({
...
     req.datatable = req.datatable || {};
+    req.datatable.schema = schema;
     return configure(req, res, next);
   };
 
   const _getApiPath = (req, res, next) => {
-    req.datatable.apiPath = apiPath;
+    if (apiPath) {
+      req.datatable.apiPath = apiPath;
+    }
```
lennym: What's the context where an `apiPath` would _not_ be set?

lennym:
I'm not sure this is doing what it says it is doing. The only functional change I can see is the disabled bit. The rest is just using an extra parameter on the request to do the same thing it was doing before, which I'd honestly avoid.

pages/common/routers/datatable.js
wheelsandcogs: `apiPath` was coming through as `undefined` and overwriting the path we actually wanted to use in some cases.

It has something to do with the way we're dynamically assigning the schema / api path in configure here:
https://github.com/UKHomeOffice/asl-internal-ui/pull/40/files#diff-8fcd7a31470eb38e82be8f8cc7f5a094R27

I can't remember the exact reason it was happening because @joefitter was helping with a few issues around the search at the same time.

wheelsandcogs:
This allows us to dynamically change the schema for a route inside the configure, which is what we need as everything in the search is under `/search?searchType=x`.

I went with `/search/people` `/search/establishments` etc initially but then there were issues with reusing the same react template for each route.

pages/common/routers/datatable.js
lennym: That's happening because the datatable router is expecting to do [this work](https://github.com/UKHomeOffice/asl-internal-ui/blob/0693623f74e441d51861dfa6939a992ca0b11924/pages/search/index.js#L37-L49) in the `getApiPath` method, rather than in configure. In fact that method exists exactly for this purpose.

lennym:
That's curious because I was under the impression that there were cases where we did exactly that (NACWOs in schedule of premises for example), but it makes sense.

👍

pages/common/routers/datatable.js
joefitter: Sorry, not sorry

PR #219, opened 2019-01-10
Admins can now apply for a named role for a user

wheelsandcogs:


pages/profile/index.js
```diff
@@ -9,7 +9,8 @@ const list = require('./list');
 const read = require('./read');
 const invite = require('./invite');
 const invitations = require('./invitations');
-const role = require('./permission');
+const permissionRouter = require('./permission');
```
lennym: Can you remove the `Router` bit for consistency with the other routers?

pages/profile/read/views/index.jsx
```diff
@@ -85,61 +91,62 @@ const Index = ({
...
+                              roles.map(({ type, id }) =>
+                                <li key={id}>{defineValue(type.toUpperCase())}</li>
+                              )
+                            }
+                          </ul>
+                        }
+                        { isEmpty(roles) &&
+                          <Snippet>responsibilities.noRoles</Snippet>
+                        }
+                        <RoleApply />
```
lennym: This should only be visible if the logged in user has permission to edit roles.

lennym: Actually, I see the component itself handles that. That's probably a better way of doing it 👍

pages/profile/roles/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+    return req.api(`/establishment/${req.establishmentId}/profile/${req.profileId}/role`, opts)
+      .then(() => res.redirect(req.buildRoute('profile.role.success')))
+      .catch(next);
+  });
+
+  app.use('/success', success({ model: 'role' }));
+
+  return app;
+};
```
lennym: There should be a permissions check somewhere in this router.

wheelsandcogs: As discussed offline, the permissions check is taken care of when this router is mounted in `pages/profile/index.js`.

pages/profile/index.js
lennym: Makes sense. Leave it be then.

PR #223, opened 2019-01-16
Display a different error message if the NTCO does not have profile read perms

wheelsandcogs:
If the NTCO cannot view all profiles at an establishment then they will not be able to endorse PILs for non-named users.

We should make this obvious using a different error message.

### DIFF
--- a/pages/task/read/index.js
+++ b/pages/task/read/index.js
@@ -6,6 +6,7 @@ const { cleanModel } = require('../../../lib/utils');
 const success = require('./routers/success');
 const getContent = require('./content');
 const { merge } = require('lodash');
+const UnauthorisedError = require('@asl/service/errors/unauthorised');
 
 module.exports = settings => {
   const app = page({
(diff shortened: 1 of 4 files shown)


pages/task/read/index.js
```diff
@@ -21,6 +22,15 @@ module.exports = settings => {
...
 
   app.use('/', (req, res, next) => {
+    Promise.resolve()
+      .then(() => req.user.can('profile.read.all', { establishment: req.task.data.data.establishmentId }))
+      .then(canReadAllProfiles => {
+        if (!canReadAllProfiles) {
+          const content = getContent(req.task);
+          next(new UnauthorisedError(content.errors.permissions));
+        }
+      });
```
lennym: This seems excessively broad-scoped. It means that anyone who can't see all profiles will _always_ be thrown an error, irrespective of whether the task is related to a profile that they cant see or not.

Specifically it will prevent any basic user or ASRU user from ever being able to process any task.

I'd suggest that the permissions check, and fallback message should only be implemented in cases where an attempt to load a profile has failed.

lennym: Leave it in the catch for now.

PR #229, opened 2019-01-18
Refactor task display to decouple from PIL endorsement

lennym:
The content of the task modules was hard-coded to a lot of PIL content, which is not relevant to other tasks.

Refactor away the PIL specific content and provide fallbacks for a generic task.

Determine the available options from the output of the workflow API, and not locally defined content, with default content values provided for the available statuses.

Moves reason field out of a conditional reveal and allows a comment to be input for all decisions, only marking it mandatory for certain decisions.

### DIFF
--- a/pages/task/read/routers/confirm.js
+++ b/pages/task/read/routers/confirm.js
@@ -13,28 +13,21 @@ module.exports = () => {
   app.use(form({
     locals: (req, res, next) => {
       const formValues = req.session.form[`${req.task.id}-decision`].values;
-      console.log(formValues);
-      const decision = formValues.decision;
 
       res.locals.static.task = req.task;
-      res.locals.static.decision = decision;
+      res.locals.static.decision = formValues.decision;
       res.locals.static.reason = formValues.reason;
       next();
     }
   }));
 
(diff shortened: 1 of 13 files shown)


pages/task/read/routers/confirm.js
```diff
@@ -13,11 +13,12 @@ module.exports = () => {
   app.use(form({
     locals: (req, res, next) => {
       const formValues = req.session.form[`${req.task.id}-decision`].values;
+      console.log(formValues);
```
wheelsandcogs: console log

pages/task/read/routers/confirm.js
```diff
@@ -32,7 +33,7 @@ module.exports = () => {
     const params = { status: stepId };
 
     if (commentRequired(stepId)) {
-      params.comment = formValues[`${stepId}-reason`];
+      params.comment = formValues.reason;
```
wheelsandcogs: if we're always allowing a comment now, do we still want this conditional?

lennym: Good catch, that cleans this section up a lot.

PR #240, opened 2019-01-29
Only show task actions if available

wheelsandcogs:


lennym:
I merged this and then realised that the better solution for handling content errors in unexpected states is probably to mark the `<Snippet optional/>` rather than adding content that _should_ never be used.

lennym:
I don't think this is worth changing, more an FYI.

PR #241, opened 2019-01-31
Add missing ntco-endorsed status

wheelsandcogs:
This might not be needed depending on the outcome of https://jira.digital.homeoffice.gov.uk/browse/AS-962


lennym:
I think it's ~ok~ good to have this. If all possible statuses can be represented in the UI then it protects us against errors.

PR #243, opened 2019-02-04
Bugfix/changes to task views

joefitter:


pages/task/read/content/base.js
```diff
@@ -1,5 +1,16 @@
...
+    title: 'Current status',
+    'with-licensing': 'With licensing',
+    'referred-to-inspector': 'Referred to inspector',
+    'returned-to-applicant': 'Returned to applicant',
+    'withdrawn-by-applicant': 'Withdrawn by applicant',
+    'with-ntco': 'With NTCO',
+    'ntco-endorsed': 'NTCO endorsed',
+    'inspector-recommended': 'Inspector recommended for approval',
+    'inspector-rejected': 'Inspector recommended for rejected'
+  },
```
lennym: I think this list now exists in a number of different places, and not necessarily in the same form. Can we consolidate?

joefitter: yes, I did have a look in asl-constants but couldn't find it

joefitter: I guess it just needs to be in the base content for all tasks, rather than asl-constants

lennym: It's not in constants, I meant various content files in `pages/task/*`

lennym: https://github.com/UKHomeOffice/asl-pages/blob/master/pages/task/read/content/success.js#L17-L41

lennym: Also here: https://github.com/UKHomeOffice/asl-pages/blob/master/pages/task/list/content/index.js#L26-L36

pages/task/content/status.js
```diff
@@ -0,0 +1,47 @@
...
+  },
+  resolved: {
+    state: 'Resolved',
+    action: 'Update licence'
+  },
+  rejected: {
+    state: 'Rejected',
+    action: 'Reject'
+  }
+};
```
wheelsandcogs: Like it 👍

lennym: I agree with @wheelsandcogs

joefitter: do either of you like it enough to merge it? :)

PR #244, opened 2019-02-04
Various bug fixes for Places restrictions

wheelsandcogs:
This should now:
 * allow you to submit a change to a place's restrictions with no other changes
 * only display the current restrictions when removing a place (AS-854)
 * only display the "changes to restrictions" field when editing a place that already has restrictions

### DIFF
--- a/pages/place/routers/amend.js
+++ b/pages/place/routers/amend.js
@@ -8,7 +8,16 @@ module.exports = settings => form(Object.assign({
   configure: (req, res, next) => {
     getSchemaWithNacwos(req, settings.schema || schema)
       .then(mappedSchema => {
-        req.model.changesToRestrictions = req.model.restrictions;
+        if (req.model.restrictions) {
+          // changesToRestrictions needs to be present in the model,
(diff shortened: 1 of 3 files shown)


pages/place/routers/amend.js
```diff
@@ -8,6 +8,16 @@ module.exports = settings => form(Object.assign({
...
       .then(mappedSchema => {
+        if (req.model.restrictions) {
+          // changesToRestrictions needs to be present in the model,
+          // otherwise no changes detected when only changing restrictions
+          req.model.changesToRestrictions = req.model.restrictions;
+        } else {
+          // only show changes field if we already have restrictions
+          delete mappedSchema.restrictions;
+          delete mappedSchema.changesToRestrictions;
+        }
```
joefitter: personally I don't like using `delete` - mainly because if mappedSchema is undefined you get a whacky useless error about not being able to convert a non object to an object. Stupid JavaScript. `_.unset` is safer

lennym: Or in that case, `mappedSchema = omit(mappedSchema, ...)`

PR #246, opened 2019-02-04
Permission notifications

wheelsandcogs:
Success notifications are now displayed when changing or removing permissions. (AS-865)

The current permission level should already be selected on the edit permissions form. (AS-864)

A breadcrumb back to the profile should now appear on the edit perms page.

### DIFF
--- a/pages/common/content/index.js
+++ b/pages/common/content/index.js
@@ -29,7 +29,8 @@ module.exports = {
           confirm: 'Confirm',
           success: 'Success'
         }
-      }
+      },
+      permission: 'Change / remove permissions'
     },
     project: {
       list: 'Projects'
--- a/pages/profile/permission/index.js
+++ b/pages/profile/permission/index.js
@@ -9,10 +9,16 @@ module.exports = settings => {
     ...settings
   });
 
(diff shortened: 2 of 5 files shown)


pages/common/content/index.js
```diff
@@ -29,7 +29,8 @@ module.exports = {
           confirm: 'Confirm',
           success: 'Success'
         }
-      }
+      },
+      permission: 'Change / remove permissions'
```
joefitter: this feels like it should be under profile

wheelsandcogs: It's a breadcrumb label - can't see any other breadcrumb content outside of this file?

pages/profile/permission/index.js
```diff
@@ -32,17 +43,20 @@ module.exports = settings => {
...
       .catch(next);
   });
 
   app.post('/remove', (req, res, next) => {
-
     const opts = {
       method: 'DELETE'
     };
 
+    res.locals.static.content = content;
```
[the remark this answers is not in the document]
wheelsandcogs: Yeah I tried that but I think it's because the route is `/remove` - i'll try adding it to the paths in page().

PR #249, opened 2019-02-05
list and read view for profile amend tasks

joefitter:


pages/task/read/routers/view.js
```diff
@@ -71,11 +71,14 @@ module.exports = () => {
...
     if (req.task.data.action === 'update' || req.task.data.action === 'delete') {
       const model = req.task.data.model;
-      const establishment = req.task.data.data.establishmentId;
+      let est = '';
+      if (model !== 'profile') {
+        est = `/establishment/${req.task.data.data.establishmentId}`;
+      }
       const id = req.task.data.id;
-      const url = `/establishment/${establishment}/${model}/${id}`;
+      const url = `/${model}/${id}`;
```
lennym: Do unscoped profile urls exist in establishment facing world? I don't think they do.

lennym: There is obviously a way round this, but I'm not sure what it is.

PR #251, opened 2019-02-05
if not autoresolved, display a more suitable message

joefitter:


pages/user/update/index.js
```diff
@@ -44,13 +44,16 @@ module.exports = settings => {
       }
     };
     return req.api(`/me`, opts)
-      .then(() => next())
+      .then(response => {
+        const status = get(response, 'json.data.json.data.status');
```
lennym: WUT?! Is that for real? We may want to... not.

joefitter: yeah, it's whack.

joefitter: json data this, json data that

lennym: How much do we want to resolve that now vs _the bucket_?

joefitter: it _works_ 😬

joefitter: but yeah I 100% agree it's completely mental. I don't think this is the only place it's used though so tempted to say let's add a bug

lennym: It's a couple of lines in the workflow client in asl-service, so I'm going to fix it. I can't handle it being the way it is and it's easier to fix now before it spreads to other places.

lennym: Turns out it isn't asl-service, it's specific to this one endpoint 

https://github.com/UKHomeOffice/asl-public-api/pull/126

lennym:
I've made the change to un-💩-ify the api response, so we can unjsondatajsondataify here now.

joefitter:
I think you mean dejsondatajsondataify

PR #256, opened 2019-02-07
bugfix - fix restrictions label in list, dont prepopulate changes to …

joefitter:
…restrictions with restrictions, allow changesToRestrictions change to be accepted by the form


lennym:
What's the actual bug that this is fixing?

joefitter:
there were a few. Firstly restrictions in SOP list was labelled "changes to restrictions", secondly we were setting the value of changesToRestrictions field from restrictions, so changesToRestrictions was always populated in task, even when no changes were made

PR #258, opened 2019-02-14
Adds view granted PIL page

wheelsandcogs:


pages/pil/read/views/index.jsx
```diff
@@ -1,31 +1,42 @@
...
+const PIL = ({ model, profile }) => {
+  return (
+    <Fragment>
+      <Header
+        title={<Snippet>title</Snippet>}
+        subtitle={profile.name}
+      />
+      { model.status === 'active'
+        ? <ModelSummary model={model} formatters={formatters} />
+        : <div><Link page="pil.update" pilId={model.id} label={<Snippet>action.applyNow</Snippet>} className="govuk-button" /></div>
```
lennym: What's the `<div>` doing here?

pages/pil/read/views/index.jsx
```diff
@@ -1,31 +1,42 @@
...
-    <Header
-      title={<Snippet>title</Snippet>}
-      subtitle={profile.name}
-    />
-    <ModelSummary formatters={formatters(url)} />
-  </Fragment>
-);
+      <div>
+        <Link page="profile.view" label={<Snippet>action.backToProfile</Snippet>} />
+      </div>
```
lennym: Same. If we want a block level wrapper then use a `<p>`.

pages/pil/read/views/index.jsx
wheelsandcogs: It's to prevent the two links showing inline.

PR #259, opened 2019-02-15
Change filter button and link label

lvnhmd:


lennym:
👍 but keep an eye out for any integration tests down the pipe that might be relying on these labels.

PR #261, opened 2019-02-18
Added project/read route, added tabs to project list table - active, …

joefitter:
…expired and drafts, added create and delete projects


pages/project/read/content/index.js
```diff
@@ -0,0 +1,6 @@
+const { merge } = require('lodash');
+const baseContent = require('../../content');
+
+module.exports = merge({}, baseContent, {
+
+});
```
lennym: Y THO?

joefitter: 😆

joefitter: it will have content, but this page only exists in my brain, which doesn't contain any content

PR #272, opened 2019-02-27
start of activity log for project tasks

joefitter:


pages/project/read/views/index.jsx
```diff
@@ -8,6 +8,19 @@ import { dateFormat } from '../../../../constants';
...
+    versions = versions.filter(v => v.createdAt >= granted.createdAt);
+  }
+  const version = versions[0];
+  return {
+    granted,
+    submitted: (version && version.status === 'submitted') ? version : null,
+    draft: (version && version.status === 'draft') ? version : null
+  };
+};
+
```
lennym: As discussed, let's move this logic to the API level.

pages/task/read/routers/view.js
```diff
@@ -23,22 +23,34 @@ const getRelevantActivity = activityLog => activityLog.filter(log => {
...
+  // redirect to project page if ?redirect=true
+  app.get('/', (req, res, next) => {
+    const model = get(req.task, 'data.model');
+    const status = get(req.task, 'status');
+    if (model === 'project' && req.query.redirect === 'true') {
+      req.versionId = status === 'resolved' ? req.project.granted.id : req.project.draft.id;
+      return res.redirect(req.buildRoute('project.version'));
+    }
+    next();
+  });
```
lennym: This is for the case where they've hit "amend and resubmit" right?

joefitter: yes, then the POST project/grant takes care of the actual resubmitting, however currently the comments box is ignored, we _might_ want to just read that from the session

joefitter: no, actually it is not, sorry I didn't actually look at the code - this actually isn't used anymore

PR #273, opened 2019-02-27
Display profile name on profile page

lvnhmd:


pages/profile/read/views/profile.jsx
```diff
@@ -37,7 +38,9 @@ class Profile extends React.Component {
 
     return (
       <Fragment>
-        {title && <h2>{title}</h2>}
+        {title && <p><h1>{title}</h1></p>}
+
+        {estName && <h3>{estName}</h3>}
```
lennym: Is this component re-used on the dashboard? If so then I don't think the profile name should be included here.

pages/profile/read/views/index.jsx
```diff
@@ -9,7 +9,7 @@ const Index = ({
 }) => {
 
   return (
-    <Profile establishment={establishment} profile={model} allowedActions={allowedActions} title={establishment.name}/>
+    <Profile title={model.name} profile={model} establishment={establishment} estName={establishment.name} allowedActions={allowedActions} />
```
lennym: I would add the profile name in an `h1` here instead of inside the `<Profile />` component.

pages/profile/read/views/profile.jsx
```diff
@@ -23,7 +21,7 @@ class Profile extends React.Component {
     } = this.props.profile;
 
     const allowedActions = this.props.allowedActions;
-    const title = this.props.title;
+    const estName = this.props.estName;
```
lennym: Does changing this prop mean it's going to be a breaking change on the dashboard, and if so should we revert to `title`?

pages/profile/read/views/index.jsx
lvnhmd: done

pages/profile/read/views/profile.jsx
lvnhmd: It is not breaking the dashboard, because I already changed it there - my thinking is, this is not a title, but the establishment name - not sure, but I will rename it

lennym: 👍

PR #287, opened 2019-03-04
Add type=submit to task confirm page buttons

lennym:
This wasn't actually submitting the forms before so no tasks could be progressed.


joefitter:
ooh - I guess they are outside the `<form>` element now then?

PR #289, opened 2019-03-04
Fix missing pil apply button

wheelsandcogs:
All users should be able to apply for a pil for themselves.

### DIFF
--- a/pages/profile/read/index.js
+++ b/pages/profile/read/index.js
@@ -6,5 +6,10 @@ module.exports = settings => {
     root: __dirname
   });
 
+  app.get('/', (req, res, next) => {
+    res.locals.static.isOwnProfile = req.user.profile.id === req.profileId;
+    next();
+  });
+
   return app;
 };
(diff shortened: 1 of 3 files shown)


pages/profile/read/index.js
```diff
@@ -6,10 +6,5 @@ module.exports = settings => {
     root: __dirname
   });
 
-  app.get('/', (req, res, next) => {
-    res.locals.static.isUser = req.user.profile.id === req.profileId;
```
lennym: Is this resolved elsewhere?

wheelsandcogs: I can't see anywhere that `isUser` gets used when viewing a profile? 

e.g. https://github.com/UKHomeOffice/asl-pages/blob/master/pages/profile/read/views/index.jsx doesn't use it

wheelsandcogs: we set `isOwnProfile` in the other PR:
https://github.com/UKHomeOffice/asl/pull/173/files

lennym: It's set on the dashboard where it's always true, but not set on the profile route where it might not be true.

I think this middleware needs to stay.

PR #290, opened 2019-03-06
Added copy for success messages, added logic in success router to asc…

joefitter:
…ertain which message to display


pages/common/routers/success.js
```diff
@@ -1,8 +1,24 @@
+const { merge, get } = require('lodash');
 const { Router } = require('express');
+const successContent = require('../content/success-messages');
 
-module.exports = ({ model = 'model' } = {}) => {
+module.exports = ({
+  model = 'model',
+  licence = 'pel',
+  type = 'amendment'
```
lennym: I'm not sure having default values here is appropriate. What's the reasoning behind them?

joefitter: most changes are pel amendments

pages/common/routers/success.js
```diff
@@ -1,8 +1,24 @@
...
+} = {}) => {
   const app = Router();
 
+  app.use((req, res, next) => {
+    if (req.model && req.model.status) {
+      type = req.model.status === 'active' ? 'amendment' : 'application';
+    }
+    const success = get(successContent, `${licence}.${type}.resubmitted`);
```
lennym: Hard-coded `resubmitted` strikes me as odd.

joefitter: yeah - it could be passed in, but it is resubmitted 100% of the time

joefitter: the logic for displaying task success messages lives elsewhere

joefitter: also resubmitted === submitted

pages/common/routers/success.js
joefitter: but yeah, I'll pass them in

PR #295, opened 2019-03-08
Bugfix/profile action permissions

joefitter:


pages/profile/read/views/profile.jsx
```diff
@@ -16,224 +16,210 @@ class Profile extends React.Component {
...
-      ({ establishmentId, status }) =>
-        status === 'active' && establishmentId === estId
-    );
-    const estRoles = roles.filter(
-      ({ establishmentId }) => establishmentId === estId
-    );
+    const activeProjects = projects.filter(({ establishmentId, status }) => status === 'active' && establishmentId === estId);
+    const estRoles = roles.filter(({ establishmentId }) => establishmentId === estId);
+
+    const canSeeProjects = activeProjects && (isOwnProfile || allowedActions.includes('project.read.all'));
```
lennym: `activeProjects` is an array, so will always resolve to true. There's no need to `&&` it here (or at least add a `.length`

joefitter: actual footage of me coding today

PR #308, opened 2019-03-25
fixed various issues and inconsistencies with nprc and pelh, also rep…

joefitter:
…laced profile.name with profile.firstName and lastName

### DIFF
--- a/pages/establishment/components/index.jsx
+++ b/pages/establishment/components/index.jsx
@@ -1,0 +1,9 @@
+import React, { Fragment } from 'react';
+import { Snippet, Link } from '@asl/components';
+
+export const ProfileLink = ({ type, profile }) => (
+  <Fragment>
+    <dt><Snippet>{type}</Snippet></dt>
+    <dd><Link page="profile.view" profileId={profile.id} label={`${profile.firstName} ${profile.lastName}`} /></dd>
+  </Fragment>
+);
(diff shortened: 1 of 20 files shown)


pages/establishment/components/index.jsx
```diff
@@ -0,0 +1,9 @@
+import React, { Fragment } from 'react';
+import { Snippet, Link } from '@asl/components';
+
+export const ProfileLink = ({ type, profile }) => (
+  <Fragment>
+    <dt><Snippet>{type}</Snippet></dt>
+    <dd><Link page="profile.view" profileId={profile.id} label={`${profile.firstName} ${profile.lastName}`} /></dd>
```
wheelsandcogs: Doesn't `profile.view` need establishmentId?
https://github.com/UKHomeOffice/asl/blob/master/lib/urls.js#L13

Or does it automatically provide it when you've got one in the URL? Just mentioning cause I had to fix a broken URL on task view (which doesn't have estId in the URL)

joefitter: it gets it from the URL. We dump req.params into static and make this available to the `Link` component. You don't need to pass any params to things that are in the scope of the current URL. You only need to provide params if linking to something else, like a profile in another establishment

joefitter: worth noting that this only works if your pages all mergeParams: 
```js
const { Router } = require('express');
```

joefitter: on a task view you probably aren't scoped to an establishment so you need to provide it

PR #309, opened 2019-03-26
Bugfix/show procedures on pil

joefitter:


pages/pil/read/index.js
```diff
@@ -16,13 +16,20 @@ module.exports = settings => {
...
       issueDate: {},
       revocationDate: {},
       conditions: {},
-      procedures: {}
+      procedures: {},
+      notesCatD: {
+        show: false
+      },
+      notesCatF: {
+        show: false
```
lennym: I think cat F notes _should_ be shown.

PR #315, opened 2019-04-03
Update index.js

pcuk1979:
Cleaning up page titles


pages/profile/roles/apply/content/index.js
```diff
@@ -2,7 +2,7 @@ const { merge } = require('lodash');
 const baseContent = require('../../../content');
 
 module.exports = merge({}, baseContent, {
-  title: 'Which role do you want to apply for?',
+  title: 'Profile amendment',
```
lennym: In the remove role page it was "Amend profile" and here it's "Profile amendment", maybe pick one.

Also the same comments as #316 apply here - there's no context with the new title.

PR #316, opened 2019-04-03
Clearing up page titles

pcuk1979:
With the proper process this time :-)

### DIFF
--- a/pages/profile/roles/remove/content/index.js
+++ b/pages/profile/roles/remove/content/index.js
@@ -5,7 +5,7 @@ module.exports = merge({}, baseContent, {
   title: 'Amend profile',
   fields: {
     role: {
-      label: ''
+      label: 'Please select the role you want to remove.'
     },
     comment: {
       label: 'Why are you removing this named role from this person?'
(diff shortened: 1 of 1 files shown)


pages/profile/roles/remove/content/index.js
```diff
@@ -2,13 +2,13 @@ const { merge } = require('lodash');
 const baseContent = require('../../../content');
 
 module.exports = merge({}, baseContent, {
-  title: 'Please select the role you want to remove',
+  title: 'Amend profile',
```
lennym: That's going to leave this page very sparse in terms of actual context. It's just going to be "Amend profile" and a list of roles with radio buttons, and no explanation of what a user is actually _doing_. Is that intentional?

pages/profile/roles/remove/content/index.js
```diff
@@ -2,13 +2,13 @@ const { merge } = require('lodash');
...
   fields: {
     role: {
       label: ''
```
lennym: We might want to add a label here in that case.

PR #317, opened 2019-04-03
Feature/as 1157 spoc and inspectors

lvnhmd:
Please do not merge as I probably need to show the **already** assigned inspectors and spocs in the sidebar

### DIFF
--- a/pages/establishment/dashboard/views/index.jsx
+++ b/pages/establishment/dashboard/views/index.jsx
@@ -72,7 +72,7 @@ const Index = ({
                 <Link key={`${asru.id}`} page="global.profile" profileId={asru.id} label={`${asru.firstName} ${asru.lastName}`} />
               ))
             }
-            { asruAdmin && <dd><Link page="establishment.asru.inspectors" label={ <Snippet>pages.edit</Snippet> } /></dd> }
+            { asruAdmin && <dd><Link page="establishment.inspectors" label={ <Snippet>pages.edit</Snippet> } /></dd> }
 
             <dt><Snippet>spoc</Snippet></dt>
             {
@@ -81,11 +81,11 @@ const Index = ({
               ))
             }
             {
-              !asruAdmin && asru.filter(p => p.asruLicensing).map(asru => (
+              asruAdmin && asru.filter(p => p.asruLicensing).map(asru => (
                 <Link key={`${asru.id}`} page="global.profile" profileId={asru.id} label={`${asru.firstName} ${asru.lastName}`} />
               ))
             }
-            { asruAdmin && <dd><Link page="establishment.asru.spocs" label={ <Snippet>pages.edit</Snippet> } /></dd> }
+            { asruAdmin && <dd><Link page="establishment.spocs" label={ <Snippet>pages.edit</Snippet> } /></dd> }
           </dl>
         </Sidebar>
       </div>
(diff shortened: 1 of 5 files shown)


pages/establishment/dashboard/views/index.jsx
```diff
@@ -30,33 +30,49 @@ const Index = ({
...
+            {
+              pelh && <ProfileLink type="pelh" profile={pelh} />
+            }
+            {
+              nprc && <ProfileLink type="nprc" profile={nprc} />
+            }
+          </dl>
+        </Sidebar>
+        <Sidebar>
+          <dl>
```
lennym: Not sure there should be two sidebars and two `<dl>`s here. Can we remove the 4 lines above here?

pages/establishment/dashboard/views/index.jsx
```diff
@@ -30,33 +30,49 @@ const Index = ({
...
-    </div>
-  </Fragment>
-);
+    </Fragment>
+  )
+  ;
+};
 
-const mapStateToProps = ({ static: { establishment, allowedActions } }) => ({ establishment, allowedActions });
+const mapStateToProps = ({ static: { establishment, allowedActions, profile } }) => ({ establishment, allowedActions, profile });
```
lennym: If the only property you need from the profile is `asruAdmin` then it makes sense to only pass that.
```suggestion
const mapStateToProps = ({ static: { establishment, allowedActions, profile } }) => ({ establishment, allowedActions, asruAdmin: profile.asruAdmin });
```

pages/establishment/dashboard/views/index.jsx
```diff
@@ -30,33 +30,49 @@ const Index = ({
...
-    <Header title={name} />
-    <div className="govuk-grid-row">
-      <div className="govuk-grid-column-two-thirds">
-        <PanelList
-          panels={links.filter(link => allowedActions.includes(link.permissions)).map(link => <DashboardLink key={link.path} { ...link } />)}
-        />
+  allowedActions,
+  profile: {
+    asruAdmin
+  }
```
lennym: Assuming you take my suggestion below, then `asruAdmin` would be a top-level prop.
```suggestion
  allowedActions,
  asruAdmin
```

pages/establishment/dashboard/views/index.jsx
lvnhmd: done

pages/establishment/dashboard/views/index.jsx
lvnhmd: done

pages/establishment/dashboard/views/index.jsx
lvnhmd: done

pages/establishment/dashboard/views/index.jsx
```diff
@@ -28,35 +28,66 @@ const Index = ({
...
-    </div>
-  </Fragment>
-);
+    </Fragment>
+  )
+  ;
+};
 
-const mapStateToProps = ({ static: { establishment, allowedActions } }) => ({ establishment, allowedActions });
+const mapStateToProps = ({ static: { establishment, allowedActions, profile } }) => ({ establishment, allowedActions, asruAdmin: profile.asruAdmin });
```
lennym: I realised this needs to also check `asruUser`
```suggestion
const mapStateToProps = ({ static: { establishment, allowedActions, profile } }) => ({ establishment, allowedActions, asruAdmin: profile.asruUser && profile.asruAdmin });
```

lvnhmd: done

PR #318, opened 2019-04-04
Move contact details / dob to top of page

wheelsandcogs:
Contact details are not a per-establishment thing, so just show them at the top.

DOB will only show if the user is viewing their own profile.

### DIFF
--- a/pages/profile/read/views/profile.jsx
+++ b/pages/profile/read/views/profile.jsx
@@ -1,6 +1,6 @@
 import React, { Fragment } from 'react';
 import isEmpty from 'lodash/isEmpty';
-import format from 'date-fns/format';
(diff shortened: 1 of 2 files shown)


pages/profile/read/views/profile.jsx
```diff
@@ -13,15 +13,13 @@ class Profile extends React.Component {
...
-      telephone,
-      email,
       roles,
       projects = [],
       establishments,
       id
     } = this.props.profile;
 
-    const allowedActions = this.props.allowedActions;
+    const allowedActions = this.props.allowedActions || [];
```
lennym: 👍 I was meaning to do this. Thanks.

PR #325, opened 2019-04-11
Add myTasks tab

lvnhmd:


pages/task/list/router.js
```diff
@@ -17,7 +17,9 @@ module.exports = ({
     res.locals.static.profileName = `${firstName} ${lastName}`;
     res.locals.static.progress = req.query.progress;
     res.locals.datatable.progress = req.query.progress;
-    res.locals.static.tabs = tabs;
+
+    const isAsru = req.user.profile.asru && req.user.profile.asru.length > 0;
```
lennym: I think you will also need to check `profile.asruUser` here.

lvnhmd: done

PR #332, opened 2019-04-15
Show withdraw application as a button instead of the radio option

wheelsandcogs:


pages/task/read/views/project.jsx
```diff
@@ -39,21 +40,29 @@ const Project = ({ task, formFields, project, establishment, children }) => {
             </Fragment>
         }
       </StickyNavAnchor>
+
       {
-        !!task.nextSteps.length && (
+        schema.status.options.length > 0 &&
```
lennym: You'll need a `!!` to prevent it rendering a `0` when it's zero.

pages/task/read/views/project.jsx
```diff
@@ -39,21 +40,29 @@ const Project = ({ task, formFields, project, establishment, children }) => {
...
+            <Form />
+            { task.canBeWithdrawn && <WithdrawApplication showHeading /> }
+          </StickyNavAnchor>
+      }
+
+      {
+        schema.status.options.length === 0 && task.canBeWithdrawn &&
+          <StickyNavAnchor id="withdraw">
+            <h2><Snippet>sticky-nav.withdraw</Snippet></h2>
+            { task.canBeWithdrawn && <WithdrawApplication /> }
```
lennym: This extra condition is redundant since it's already been applied 3 lines up ☝️

lennym:
This feels very repetitive. You seem to have added the same content to each of the model type-specific views. Would it be better to add this once at the level above?

joefitter:
I'm not sure about this, it feels like we are solving one single case with this, what about when there is only one option and it isn't withdraw? How does it look when there are radio buttons for some actions and a button just for withdraw? The design on the ticket only shows the case where there is one option - withdraw

pages/task/read/views/project.jsx
wheelsandcogs: Good point, this is from moving stuff in and out of the conditions many times to try and get the logic right.

wheelsandcogs:
@lennym it needs to be inside the `StickyNavPage` if we want it to appear in the nav, which would make it a PitA to have in the level above.

wheelsandcogs:
We could extract it to a component to avoid the repetition, but PIL has customisations and no doubt others will as well.

lennym:
Fair enough.

wheelsandcogs:
@joefitter I tried a more generic approach at first but was struggling to get it to work, so I gave up - if you fancy a crack at it, feel free to push some changes 😃

joefitter:
I don't necessarily have too much beef with the approach, more just if the implementation covers all cases - what if there is another case where a single radio button appears and it isn't withdraw?

wheelsandcogs:
We've still got the edge case of a single radio but I've created a separate ticket for that: AS-1245

pages/task/read/views/place.jsx
```diff
@@ -100,19 +102,36 @@ const Playback = ({ task, values = {}, model, establishment, formFields, isAsru,
...
+const mapStateToProps = ({
+  model,
+  static: { values, establishment, isAsru, schema: decisionSchema }
+}) => ({
+  model,
+  values,
+  establishment,
+  isAsru,
+  decisionSchema
+});
```
lennym: #jode

PR #333, opened 2019-04-15
Display none when no assigned asru

lvnhmd:


lennym:
I think the markup around here is getting a little unruly. In particular, I think the inspector/spoc part should be pulled out into a separate function to render that, ideally with consistent use of `dt`/`dd` elements.

Out of scope for this PR though.

PR #341, opened 2019-04-23
Licence status banner

wheelsandcogs:


pages/establishment/dashboard/views/index.jsx
```diff
@@ -24,6 +26,7 @@ const DashboardLink = ({ path }) => (
 );
 
 const Index = ({
+  establishment,
   establishment: {
```
joefitter: you have 2 keys referring to the same thing here, I'd either pick off the establishment properties you need for the banner, or get rid of the destructuring and use `establishment.property` wherever we need it

pages/pil/read/index.js
```diff
@@ -8,7 +8,7 @@ module.exports = settings => {
   });
 
   app.get('/', (req, res, next) => {
-    if (req.pil.status !== 'active') {
+    if (req.pil.status === 'pending') {
```
joefitter: is `pending` the status we give a PIL when it's created but not submitted? I thought we usually defaulted to inactive

lennym:
Does the banner have a default date format? If so then I'd not bother passing it all the time.

PR #342, opened 2019-04-24
Add traverse function

lvnhmd:


lib/utils/index.js
```diff
@@ -94,6 +94,35 @@ const normalise = str => {
   return str.toLowerCase().split(' ').join('-').replace(/[^a-z0-9]+/g, '');
 };
 
+const traverse = (node, key, keys) => {
+  if (key) keys.push(key);
```
joefitter: curly's please :)

pages/project-version/middleware/index.js
```diff
@@ -110,8 +113,51 @@ const canComment = () => (req, res, next) => {
     .catch(next);
 };
 
+const getPreviousVersion = () => (req, res, next) => {
+
+  const versions = orderBy(req.project.versions.filter(pv => pv.id !== req.versionId),
+    v => moment(v.createdAt).format('YYYY-MM-DD'), ['desc']);
```
joefitter: `desc` doesn't need to be an array here

joefitter: also I'm pretty sure you can just orderBy the UTC string without having to call moment and format here

pages/project-version/middleware/index.js
```diff
@@ -110,8 +113,51 @@ const canComment = () => (req, res, next) => {
...
+
+  if (versions.length > 0) {
```
joefitter: what happens when this check fails? We don't seem to be calling next so looks like it will hang

pages/project-version/middleware/index.js
```diff
@@ -110,8 +113,51 @@ const canComment = () => (req, res, next) => {
...
+    let prevVersion = versions.find(v => {
+      if (moment(req.version.createdAt).isAfter(v.createdAt)) {
```
joefitter: you don't need an `if` here - just return the condition

pages/project-version/middleware/index.js
```diff
@@ -110,8 +113,51 @@ const canComment = () => (req, res, next) => {
...
+      })
+      .then(() => next())
+      .catch(next);
+  }
+
+};
+
+const getVersionChanges = () => (req, res, next) => {
+  if (req.prevVersion) {
+    const cvKeys = traverse(req.version.data, null, []);
```
joefitter: I would make the second and third arguments in `traverse` optional - then you won't need to provide them first time round

pages/project-version/middleware/index.js
```diff
@@ -110,8 +113,51 @@ const canComment = () => (req, res, next) => {
...
+    const removed = remove(pvKeys, k => !cvKeys.includes(k));
+    let changed = [];
+    cvKeys.forEach(k => {
+      let cvNode = getNode(req.version.data, k);
+      let pvNode = getNode(req.prevVersion.data, k);
+      if (!isEqual(cvNode, pvNode)) {
+        changed.push(k);
+      }
+    });
+    res.locals.static.changed = added.concat(removed).concat(changed);
```
joefitter: you'll also need to pass this to the client - have a look at `./js/index.js` - we pass the initial state when starting the app

pages/project-version/middleware/index.js
```diff
@@ -1,4 +1,7 @@
 const { chain, get } = require('lodash');
+const { orderBy, remove, isEqual } = require('lodash');
+const moment = require('moment');
+const { traverse, getNode } = require('../../../lib/utils');
```
joefitter: these functions are specific to this implementation - let's not hide them in utils

joefitter:
this mostly looks fine, however you will need to pass them to the client - see my comment regarding this. Also does this handle custom fields like duration, species etc?

lib/utils/index.js
lvnhmd: done

pages/project-version/middleware/index.js
lvnhmd: done

pages/project-version/middleware/index.js
lvnhmd: done

pages/project-version/middleware/index.js
lvnhmd: done

pages/project-version/middleware/index.js
lvnhmd: done

pages/project-version/middleware/index.js
lvnhmd: done

pages/project-version/middleware/index.js
lvnhmd: done

pages/project-version/middleware/index.js
```diff
@@ -113,32 +112,58 @@ const canComment = () => (req, res, next) => {
...
 
     req.api(`/establishments/${req.establishmentId}/projects/${req.projectId}/project-versions/${prevVersion.id}`)
       .then(({ json: { data } }) => {
         req.prevVersion = data;
       })
       .then(() => next())
       .catch(next);
-  }
-
+  } else { next(); }
```
joefitter:
```js
if () {
  ...
} else {
  next();
}
```
pwease

joefitter: seeing it on one line hurts my eyes

pages/project-version/middleware/index.js
```diff
@@ -113,32 +112,58 @@ const canComment = () => (req, res, next) => {
     .catch(next);
 };
 
+const traverse = (node, ...params) => {
+  const key = (params.length > 0) ? params.shift() : null;
+  let keys = (params.length > 0) ? params.shift() : [];
```
joefitter: I was thinking just using default args... 
```js
const traverse = (node, key, keys = [])
```

pages/project-version/read/js/index.js
```diff
@@ -20,6 +20,7 @@ start({
     user: `${state.static.user.firstName} ${state.static.user.lastName}`,
     basename: state.static.basename,
     schemaVersion: state.model.project.schemaVersion,
-    establishment: state.static.establishment.name
+    establishment: state.static.establishment.name,
+    changed: state.static.changed
```
joefitter: we might want to pull this out to the top level (as per comments) as it's not going to be a static property - we're going to be manipulating the array client side as and when changes are made - it will need a reducer and some actions in asl-projects

pages/project-version/middleware/index.js
lvnhmd: done

pages/project-version/middleware/index.js
lvnhmd: done

pages/project-version/read/js/index.js
lvnhmd: done

PR #343, opened 2019-04-24
Add a toggle to activity log to hide all but most recent

wheelsandcogs:


pages/common/assets/sass/style.scss
```diff
@@ -54,20 +54,26 @@
   }
 }
 
+#activity .toggle-switch {
```
lennym: I'd generally try to avoid using id selectors in CSS ever as they're super specific and so become really difficult to extend, which leads to slightly hateful code in future.

pages/task/read/views/activity-log.jsx
```diff
@@ -17,36 +31,84 @@ const getStatusBadge = eventName => {
...
-            <p>{format(log.createdAt, dateFormat.datetime)}</p>
-          </li>
-        ))}
-      </ul>
-    </Fragment>
+    <li key={log.id}>
+      <span className="date">{format(log.createdAt, dateFormat.datetime)}</span>
+      {getStatusBadge(log.eventName)}
+      {getAuthor(log.changedBy, task)}
+      <p className="comment">{log.comment}</p>
```
lennym: Out of scope here as I know this is just reformatting, but this should be `<Markdown/>` and not `<p>`

pages/task/read/views/activity-log.jsx
```diff
@@ -17,36 +31,84 @@ const getStatusBadge = eventName => {
...
   );
 };
 
+class ActivityLog extends Component {
+  constructor(props) {
+    super(props);
+
+    this.task = props.task;
+    this.ExtraMeta = props.ExtraMeta;
+    this.latestActivity = props.task.activityLog.shift();
```
lennym: I don't think doing this work in the constructor is correct. If things are props they should just be used directly as props in `render`, and any work to modify them should be done at render-time.

Performing the work at a constructor level means that if the props are mutated then the component will not re-render.

pages/common/assets/sass/style.scss
wheelsandcogs: I need to only select the toggle switch on this page, rather than all toggle-switches - should I just make the class something like `activity-toggle-switch` ?

pages/task/read/views/activity-log.jsx
wheelsandcogs: I don't think we use rich text editors for comments yet do we?

pages/task/read/views/activity-log.jsx
wheelsandcogs: The issue here is the `this.latestActivity = props.task.activityLog.shift();` - if you do that in the render then every time you open and close the activity you lose the first entry until there's nothing left.

lennym: Then don't use `.shift()` - take the first element by doing `activityLog[0]`, and when rendering the rest check `index > 0`.

pages/common/assets/sass/style.scss
lennym: Just target the container with a classname instead of the id. You might need to replace the wrapping `<Fragment>` below with a `<div className="activity-log">` to do this.

PR #349, opened 2019-04-29
Highlight amends in grey

lvnhmd:


pages/project-version/middleware/index.js
```diff
@@ -164,21 +164,57 @@ const getPreviousVersion = () => (req, res, next) => {
...
+const getGrantedChanges = () => (req, res, next) => {
+  if (req.grantedVersion) {
+    let granted = getChanges(req.version, req.grantedVersion);
+    if (res.locals.static.changes.latest) {
+      granted = granted.filter(e => !res.locals.static.changes.latest.includes(e));
+    }
+    res.locals.static.changes = {
+      ...res.locals.static.changes,
+      granted
+    };
```
joefitter: `getGrantedChanges` should not depend on `getLatestChanges` - if these are called out of turn you'll end up trying to call an array spread on `undefined` - We should make getLatestChanges and getGranted changes just plain old functions which take the request, and resolve with the required changes, we'd then just export a middleware which adds the changes to the response - something like:

```js
const getAllChanges = (req, res, next) => {
  Promise.all([
    getLatestChanges(req),
    getGrantedChanges(req)
  ])
      res.locals.static.changes = { latest, granted };
    })
    .then(() => next())
    .catch(next);
}
```

lvnhmd: done

pages/project-version/middleware/index.js
```diff
@@ -164,29 +164,60 @@ const getPreviousVersion = () => (req, res, next) => {
...
+    let pvNode = getNode(version.data, k);
+    if (!isEqual(cvNode, pvNode)) {
+      changed.push(k);
+    }
+  });
+  return added.concat(removed).concat(changed);
+};
+
+const getAllChanges = () => (req, res, next) => {
+  console.log('getAllChanges');
```
joefitter: remove me pls

PR #351, opened 2019-04-30
Bugfix/clear session on submit

joefitter:


lennym:
Looks legit. Do we want this in before S&T?

joefitter:
probs not

joefitter:
it has the potential to break things - I have tested it a lot, but let's just wait until after :)

PR #352, opened 2019-05-01
Don't render activity log if there's no activity

wheelsandcogs:
Activity log prop might exist but be empty.


lennym:
I'm not sure I agree with this.

It's a theoretically impossible position for a task to both exist and have no activity - it should always have at least the activity record corresponding to its creation.

And we shouldn't attempt to render impossible states - throwing an error is correct in this case.

PR #358, opened 2019-05-07
Drafts don't have granted versions

wheelsandcogs:


pages/project-version/middleware/index.js
```diff
@@ -194,7 +198,7 @@ const getAllChanges = () => (req, res, next) => {
     .then(([latest, granted]) => {
       res.locals.static.changes = {
         latest,
-        granted: granted.filter(e => !latest.includes(e))
+        granted: (granted || []).filter(e => !latest.includes(e))
```
joefitter: this filter shouldn't be here at all - the client should be taking care of what to show as granted - granted in this case is incorrect as it only shows a subset of granted changes. But also I don't care enough to fix it.

lennym: I agree with Joe. The server is over-reaching here and making render-time decisions in data processing.

It's also out of the scope of this PR.

PR #359, opened 2019-05-07
Add deadline extension to activity log

wheelsandcogs:


pages/task/read/routers/view.js
```diff
@@ -8,7 +8,12 @@ const { cleanModel } = require('../../../../lib/utils');
 const getContent = require('../content');
 const { getNacwoById, getEstablishment } = require('../../../common/helpers');
 
-const getRelevantActivity = activityLog => activityLog.filter(log => {
+const getRelevantActivity = activityLog => activityLog.filter((log, index) => {
+  if (get(log, 'event.meta.payload.data.extended') && index > 0) {
+    log.eventName = 'status:deadline-extension';
```
wheelsandcogs: I've overwritten the event name here (it would be a generic 'update') with a made up status so that we can reuse the ExtraMeta component with the fewest amount of changes. Not sure how you guys will feel about that?

pages/task/read/views/index.jsx
```diff
@@ -14,20 +14,41 @@ import Profile from './profile';
 import Role from './role';
 import Project from './project';
 import get from 'lodash/get';
+import cloneDeep from 'lodash/cloneDeep';
 import { dateFormat } from '../../../../constants';
 import format from 'date-fns/format';
 import parse from 'date-fns/parse';
+import moment from 'moment-business-time';
```
wheelsandcogs: I've imported `moment` here but we're already using `date-fns`. I've not seen a similar business time plugin for it but there is this comment: https://github.com/date-fns/date-fns/issues/584#issuecomment-487078092 which might get us most of the way there?

wheelsandcogs: Same is true in the deadline component that was already merged.

pages/task/read/routers/view.js
```diff
@@ -8,7 +8,12 @@ const { cleanModel } = require('../../../../lib/utils');
 const getContent = require('../content');
 const { getNacwoById, getEstablishment } = require('../../../common/helpers');
 
-const getRelevantActivity = activityLog => activityLog.filter(log => {
+const getRelevantActivity = activityLog => activityLog.filter((log, index) => {
+  if (get(log, 'event.meta.payload.data.extended') && index > 0) {
```
joefitter: why are we checking index > 0? I'd either check for something more explicit or add a comment

wheelsandcogs: If extending the deadline was the most recent activity then we don't show it in the activity log, we only show it in the log once the task status changes.

lennym: I think that's an old requirement that we got rid of. See https://jira.digital.homeoffice.gov.uk/browse/AS-1231

pages/task/read/routers/view.js
lennym: Fine.

pages/task/read/routers/view.js
wheelsandcogs: Ah, the design and the ticket had conflicting info - I'll remove the check.

wheelsandcogs: Sorted.

PR #362, opened 2019-05-10
Fix spacing in inspector and spoc lists

wheelsandcogs:


pages/establishment/dashboard/views/index.jsx
```diff
@@ -63,7 +63,10 @@ const Index = ({
             }
             {
               asruAdmin && establishment.asru.filter(p => p.asruInspector).map(asru => (
-                <Link key={`${asru.id}`} page="global.profile" profileId={asru.id} label={`${asru.firstName} ${asru.lastName}`} />
+                <Fragment key={`${asru.id}`}>
+                  <Link page="global.profile" profileId={asru.id} label={`${asru.firstName} ${asru.lastName}`} />
+                  <br />
+                </Fragment>
```
joefitter: hate to be _that guy_ but seems like an unnecessary use of `<Fragments>` and `<br />`'s - can't we wrap them in <p> tags?

joefitter: if wrapping it in a `p` doesn't look right I'd prefer to just add some css to the Link elements

joefitter: ooh - my <p> was rendered as HTML

PR #366, opened 2019-05-14
lookup correct previous version. If there is a granted version, only …

joefitter:
…lookup versions that are either this version, or a more recent draft


pages/project-version/middleware/index.js
```diff
@@ -198,7 +207,7 @@ const getAllChanges = () => (req, res, next) => {
     .then(([latest, granted]) => {
       res.locals.static.changes = {
         latest,
-        granted: (granted || []).filter(e => !latest.includes(e))
+        granted: (granted || []).filter(e => !(latest || []).includes(e))
```
lennym: 🤮

PR #367, opened 2019-05-15
Add licence status banner to pdf when not active

wheelsandcogs:


pages/project-version/pdf/index.jsx
```diff
@@ -32,7 +34,7 @@ module.exports = settings => {
           headerTemplate: header,
           footerTemplate: footer,
           margin: {
-            top: 100,
+            top: hasStatusBanner() ? 180 : 100,
```
joefitter: I can get on board with saving the result of the condition to a single-use variable, but having a function which always returns the same value seems particularly redundant

pages/project-version/pdf/views/header.jsx
```diff
@@ -1,12 +1,14 @@
...
+import StatusBanner from './status-banner';
 
 const format = date => moment(date).format('DD MMM YY');
 
-const Header = ({ project }) => (
-  <Wrapper name="header">
+const Header = ({ project, nonce }) => (
+  <Wrapper name="header" nonce={nonce}>
     <header>
+      <StatusBanner licence={project} licenceType="ppl" />
```
joefitter: Do we _always_ want a status banner? If so is the previous conditional with different padding now redundant?

joefitter: OIC, so the StatusBanner returns `null` if active - as you were :)

joefitter:
looks fine, one minor stylistic whinge, but other than that g2g

pages/project-version/pdf/index.jsx
wheelsandcogs: Yeah had a bit of a brain fart there, not sure why I used a function.

PR #370, opened 2019-05-16
Use the model from the task rather than attempting to fetch a potentially deleted model

wheelsandcogs:
Requires https://github.com/UKHomeOffice/asl-workflow/pull/97 to be merged first.

### DIFF
--- a/pages/task/read/routers/view.js
+++ b/pages/task/read/routers/view.js
@@ -116,26 +116,9 @@ module.exports = () => {
         res.locals.static.values = req.user.profile;
         return next();
       }
-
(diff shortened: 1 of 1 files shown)


pages/task/read/routers/view.js
```diff
@@ -117,6 +117,11 @@ module.exports = () => {
         return next();
       }
 
+      if (req.task.data.action === 'delete') {
+        res.locals.static.values = req.task.data.deleted;
```
lennym: I think if you follow my suggestion on the workflow PR then this is not needed.

PR #372, opened 2019-05-17
Changes to task statuses and actions

pcuk1979:
Not many changes here.

It would be good to have an additional  action of 'Amend licence' for the Approved state for amendments.

I'm also unsure what Incomplete, Case opened and Case updated are. Will these appear as labels in a task list?

### DIFF
--- a/pages/pil/exemptions/content/modules.js
+++ b/pages/pil/exemptions/content/modules.js
@@ -17,7 +17,7 @@ module.exports = merge({}, content, {
       required: 'You need to select at least one module.'
     },
     reason: {
-      customValidate: 'Please enter a reason that an applicant should be exempt from module {{meta}}.'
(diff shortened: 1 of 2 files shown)


lennym:
Could you create a ticket for the "Amend licence" thing if there isn't already one.

The things you're not sure about are generally activity types which theoretically _can_ exist in the activity log, but are filtered out at a UI level. We define labels for them in order to not crash the system if we make a mistake with the filtering code elsewhere and they stop being filtered. I think it's fine to leave them as they are.

pcuk1979:
Cheers

There's a ticket for Decision boxes here

https://jira.digital.homeoffice.gov.uk/browse/AS-1350

The Amend licence is the only additional status needed, but the accompanying subtexts for each option makes a greater distinction between applications and amendments - however, I've tried to reduce the number of differences wherever possible.

PR #383, opened 2019-05-23
updated task view to remove the duplicate forms, now withdraw is a qu…

joefitter:
…ery action, and other fields (restricitons) can be included in the body

### DIFF
--- a/pages/task/read/views/make-decision.jsx
+++ b/pages/task/read/views/make-decision.jsx
@@ -1,35 +1,16 @@
 import React, { Fragment } from 'react';
-import { connect } from 'react-redux';
-import { Form } from '@asl/components';
+import { Button } from '@ukhomeoffice/react-components';
 
-const MakeDecision = ({ schema, csrfToken }) => {
-  return (
-    <Fragment>
-      {
-        // if there is only a single option (excluding withdraw) then display it as a button
-        schema.status.options.length === 1 &&
-        <form method="POST" noValidate>
-          <button className="govuk-button" type="submit">{schema.status.options[0].label}</button>
-          <input type="hidden" name="_csrf" value={csrfToken} />
-          <input type="hidden" id={`status-${schema.status.options[0].value}`} name="status" value={schema.status.options[0].value} />
-        </form>
-      }
+const MakeDecision = ({ schema, formFields }) => {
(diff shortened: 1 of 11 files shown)


pages/task/read/views/make-decision.jsx
```diff
@@ -1,35 +1,16 @@
...
-        </form>
-      }
+const MakeDecision = ({ schema, formFields }) => {
+  return schema.status.options.length > 1
+    ? formFields
+    : (
+      <Fragment>
+        <input type="hidden" id={`status-${schema.status.options[0].value}`} name="status" value={schema.status.options[0].value} />
+        <Button>{schema.status.options[0].label}</Button>
+      </Fragment>
```
wheelsandcogs: The reason this wasn't a ternary was because it needs to render nothing if there are zero options (e.g. a completed task). Not sure what this will do in that case (error on missing index 0?)

pages/task/read/views/pil.jsx
```diff
@@ -153,8 +153,12 @@ const Pil = ({ profile, task, children, schema }) => {
         schema.status.options.length > 0 &&
           <StickyNavAnchor id={getNtcoStatus(task.status)}>
             <h2><Snippet>{`sticky-nav.${getNtcoStatus(task.status)}`}</Snippet></h2>
-            <MakeDecision />
-            { task.canBeWithdrawn && <WithdrawApplication showHeading /> }
+            {
+              <MakeDecision schema={schema} formFields={formFields} />
+            }
```
wheelsandcogs: These brackets seem redundant.

pages/task/read/views/make-decision.jsx
wheelsandcogs: It looks like it doesn't get called except when inside a check for options > 0, but we should probably have a check in the component itself.

joefitter: nah, one check is good enough for me :) the reason I removed it is because it is only ever included when there _are_ options, I'd go with one or the other, always render and check in the component, or only render when applic. The latter resulting in less code running

pages/task/read/views/pil.jsx
joefitter: yes, yes they are, good spot

PR #388, opened 2019-05-24
Add input field for other accrediting body + playback

wheelsandcogs:


pages/pil/training/schema/certificate.js
```diff
@@ -9,8 +25,8 @@ module.exports = {
     ]
   },
   accreditingBody: {
-    inputType: 'select',
-    options: accreditingBodies,
+    inputType: 'radioGroup',
+    options: accreditingBodyOptions(),
```
joefitter: this is purely me being a pedant, but I really don't like impure functions - if the result of the function needs accreditingBodies, then these should be passed in as an argument, or better still, map them inline here :D

lennym:
The logic for display seems to default to `otherAccreditingBody` when it is set, but we don't _seem_ to unset it anywhere if it's populated, but then the radio selection changes.

We probably want to do something like `values.accreditingBody === 'Other' ? values.otherAccreditingBody : values.accreditingBody` to make sure we check the value of the radio button first.

PR #399, opened 2019-06-04
Feature/establishment conditions

joefitter:


pages/establishment/details/views/index.jsx
```diff
@@ -56,14 +57,22 @@ const Index = ({
...
-                  <Fragment>
-                    <p><Snippet>conditions.hasConditions</Snippet></p>
-                    <ReactMarkdown>{ establishment.conditions }</ReactMarkdown>
-                  </Fragment>
-                )
-                : <p><Snippet>conditions.noConditions</Snippet></p>
+              {
+                <Conditions
+                  conditions={establishment.conditions}
+                  canUpdate={asruUser && !openTask}
```
lennym: Can we make this dependent on a permissions check or an `allowedActions` or something similar? I'd generally prefer to avoid spreading the logic of _why_ a person can do a thing around the codebase as much as possible.

joefitter: sure, I think it should be in allowedActions

pages/establishment/details/index.js
```diff
@@ -1,15 +1,44 @@
...
   const app = page({
     ...settings,
     root: __dirname
   });
 
-  app.get('/', (req, res, next) => {
+  app.use((req, res, next) => {
     req.breadcrumb('establishment.read');
+    req.model = req.establishment;
+    res.locals.static.asruUser = req.user.profile.asruUser;
```
lennym: This _shouldn't_ be needed anymore if you've switched to `allowedActions`.

joefitter: yes good spot 😅

PR #400, opened 2019-06-04
Remove unused property

lvnhmd:


lennym:
`isGrantedVersion` needs to be removed from https://github.com/UKHomeOffice/asl-pages/pull/400/files#diff-c2ddb31c48999a820a51d581c96d8982R23 too.

PR #403, opened 2019-06-06
Add animal type to exemptions

lvnhmd:


pages/pil/update/views/index.jsx
```diff
@@ -59,6 +59,9 @@ const Index = ({ establishment, certificates, exemptions, model, skipExemptions,
       models: exemptions,
       schema: {
         module: {},
+        species: {
+          format: species => (species && species.length > 0 && <span className="species"> ({species.join(', ')})</span>)
```
joefitter: extra `()`'s not needed

pages/pil/exemptions/routers/modules.js
```diff
@@ -74,7 +99,8 @@ module.exports = () => {
               json: {
                 data: {
                   module,
-                  description: req.form.values[`module-${module}-reason`]
+                  description: req.form.values[`module-${module}-reason`],
+                  species: castArray(req.form.values[`module-${module}-species`]).filter(s => s !== '')
```
joefitter: why is this filter here? do we have empty string animals being added somehow?

pages/pil/update/views/index.jsx
lvnhmd: that has changed now

pages/pil/exemptions/routers/modules.js
lvnhmd: because one can click on 'Add another', not chose species and then submit, in this case there will be one empty string it the array of species

pages/pil/exemptions/routers/modules.js
```diff
@@ -1,42 +1,68 @@
...
+          }
+
+          return type;
+        }, {})
+      };
+      next();
     },
     locals: (req, res, next) => {
-      res.locals.static.schema = schema;
+      res.locals.static.schema = omit(req.form.schema, moduleCodes.map(c => `module-${c}-reason`).concat(moduleCodes.map(c => `module-${c}-species`)));
```
lennym: I have no idea what this line of code is doing apart from "too much". Can we split it out a little to make it clearer?

pages/pil/exemptions/routers/modules.js
```diff
@@ -1,42 +1,68 @@
...
-          ...obj,
-          [`module-${val.value}-reason`]: val.reveal.reason
-        };
-      }, {})
+    configure: (req, res, next) => {
+      req.form.schema = {
+        ...schema,
+        ...schema.modules.options.reduce((obj, val) => {
+
+          let type = {
```
lennym: This should be `const`.

pages/pil/exemptions/routers/modules.js
```diff
@@ -1,42 +1,69 @@
...
 const { moduleCodes } = require('@asl/constants');
+const { species } = require('@asl/constants');
+const content = require('../content/modules');
+
+const modulesThatRequireSpecies = [
+  'PILA (theory)',
+  'PILA (skills)',
+  'K (theory)',
+  'K (skills)'
+];
```
lennym: This list is also defined at https://github.com/UKHomeOffice/asl-pages/blob/master/pages/pil/training/routers/modules.js#L7-L12 - it might be worth pulling it out to a shared file (probably still under `pages/pil` though, not all the way into constants)

pages/pil/exemptions/routers/modules.js
lvnhmd: done

pages/pil/exemptions/routers/modules.js
lvnhmd: done

PR #408, opened 2019-06-07
Pass version to licence status banner component

lennym:
This encapsulates all of the logic for whether to render a banner or not inside the banner component itself rather than including it in each page. This brings the project version instance inline with the other pages which do not include custom logic in the page.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "@asl/components": {
-      "version": "3.17.4",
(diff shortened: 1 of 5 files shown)


wheelsandcogs:
Do you wanna bump components here before I merge?

lennym:
Yes. I wrote it to be not-breaky if I didn't, but it'll be less worky without it.

wheelsandcogs:
Alright, there's no need to show off with your technical terms! 😉

PR #409, opened 2019-06-10
Bugfix/success pages

joefitter:
* Use pel for establishment, role and place updates
 * dont clear status update from session, so success page can be reloaded
 * model status fixed in workflow - now status when task submitted

### DIFF
--- a/pages/task/read/routers/success.js
+++ b/pages/task/read/routers/success.js
@@ -19,7 +19,7 @@ module.exports = () => {
   });
 
   app.use((req, res, next) => {
-    const model = req.task.data.model === 'place' || req.task.data.model === 'role' || req.task.data.model === 'establishment'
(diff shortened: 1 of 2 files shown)


pages/task/read/routers/success.js
```diff
@@ -14,22 +14,22 @@ module.exports = () => {
...
     const status = get(req.session, `form.${req.model.id}.values.status`);
     req.status = status;
-
     res.locals.static.profile = req.task.data.changedBy;
     next();
   });
 
   app.use((req, res, next) => {
-    const model = req.task.data.model === 'place' || req.task.data.model === 'role' ? 'pel' : req.task.data.model;
+    const model = req.task.data.model === 'place' || req.task.data.model === 'role' || req.task.data.model === 'establishment'
```
lennym:
```suggestion
    const model = ['place', 'role', 'establishment'].includes(req.task.data.model)
```

PR #417, opened 2019-06-17
Add amend establishment feature

wheelsandcogs:


pages/establishment/formatters/flag-grouping.js
```diff
@@ -0,0 +1,31 @@
...
+        establishment.licences.push(licence);
+      }
+    });
+
+    establishment.authorisations = [];
+    ['killing', 'rehomes'].map(authorisation => {
+      if (establishment[authorisation] === true) {
+        establishment.authorisations.push(authorisation);
+      }
+    });
```
lennym: These `.map.push` arrangements can be much simplified by using `.filter` instead.

```js
establishment.licences =  ['supplying', 'breeding', 'procedure'].filter(licence => establishment[licence]);
establishment.authorisations =  ['killing', 'rehomes'].filter(licence => establishment[licence]);
```

pages/establishment/update/index.js
```diff
@@ -0,0 +1,80 @@
...
+  });
+
+  app.use('/confirm', confirm());
+
+  app.post('/confirm', (req, res, next) => {
+    let values = ungroupFlags(req.session.form[req.model.id].values);
+    const { comments } = values;
+
+    const params = {
+      data: omit(values, ['licences', 'authorisations', 'comments']),
```
lennym: Not a huge deal, but I'd _probably_ prefer a `pick` of the known good fields than an `omit`, it's a little safer.

pages/establishment/update/index.js
```diff
@@ -0,0 +1,80 @@
...
+        validate: ['required']
+      }
+    }),
+    process: (req, res, next) => {
+      if (!Array.isArray(req.body.licences)) {
+        req.form.values.licences = req.body.licences ? [req.body.licences] : [];
+      }
+      if (!Array.isArray(req.body.authorisations)) {
+        req.form.values.authorisations = req.body.authorisations ? [req.body.authorisations] : [];
+      }
```
lennym: I'm not sure I follow what happens here if it _is_ an array.

pages/place/routers/confirm.js
```diff
@@ -35,6 +35,10 @@ module.exports = settings => form(Object.assign({
     }
     return res.redirect(req.originalUrl.replace(/\/confirm/, ''));
   },
+  editAnswers: (req, res, next) => {
+    delete req.session.form[req.model.id].validationErrors;
+    return res.redirect(req.baseUrl.replace(/\/confirm/, ''));
+  },
```
lennym: What was this doing before? Straight-up not working?

pages/establishment/formatters/flag-grouping.js
```diff
@@ -0,0 +1,31 @@
...
+  },
+
+  ungroupFlags: establishment => {
+    ['supplying', 'breeding', 'procedure'].map(licence => {
+      establishment[licence] = establishment.licences.includes(licence);
+    });
+
+    ['killing', 'rehomes'].map(authorisation => {
+      establishment[authorisation] = establishment.authorisations.includes(authorisation);
+    });
```
lennym: As above with `.filter`

pages/establishment/update/content/fields.js
```diff
@@ -0,0 +1,26 @@
...
+    label: 'Authorisations',
+    options: {
+      killing: 'Methods of killing not specified in Schedule 1',
+      rehomes: 'Setting free and rehoming of protected animals after completion of a series of regulated procedures'
+    }
+  },
+  comments: {
+    label: 'Why are you making this change?'
+  }
+};
```
lennym: A lot fo the content in this file feels like it's being repeated, and possibly should be shared in a common file.

pages/establishment/formatters/index.jsx
```diff
@@ -0,0 +1,18 @@
+import { capitalize } from 'lodash';
+
+const listFormatter = list => {
+  if (list.length < 1) {
+    return 'None selected';
+  }
+
+  return list.map(item => item === 'procedure' ? 'Procedures' : capitalize(item)).join(', ');
```
lennym: Where is this used? It seems odd to not use the full wording for these options.

wheelsandcogs: These are used in the side-by-side diff, so the full wording is awkward.

pages/establishment/update/index.js
wheelsandcogs: I think this was an issue with the POSTed form values coming through as a string if a single checkbox was selected, but as an array if multiple were selected.

pages/place/routers/confirm.js
wheelsandcogs: If you submit the confirm page without checking the declaration, then you correctly get a form error, but if you then navigate back to the previous form, the declaration error persists but the page doesn't have the content for the error snippet, so it fails to render.

pages/establishment/update/index.js
lennym: I get that, this just doesn't seem to handle the case where it _is_ an array.

pages/place/routers/confirm.js
lennym: Good catch!

pages/establishment/update/index.js
wheelsandcogs: That's already handled by the form router IIRC - this just overwrites the saved form values in the cases where the form returns a string or null. I think these cases are usually handled by the form router but because these are pseudo-fields and not actual properties of the establishment model, it doesn't seem to work without this code.

PR #418, opened 2019-06-18
Set species to model

lvnhmd:


pages/pil/update/index.js
```diff
@@ -33,10 +35,17 @@ module.exports = settings => {
...
+
+    const trainingSpecies = flattenDeep(req.model.certificates.map(c => c.modules.map(m => m.species)));
+    const exemptionsSpecies = flattenDeep(req.model.exemptions.map(e => e.species));
+    const species = trainingSpecies.concat(exemptionsSpecies);
+
     req.model = {
       ...req.pil,
-      ...values
+      ...values,
+      species: species.filter((v, i) => species.indexOf(v) === i)
```
lennym: Is this effectively a uniqueness test? If so could we use something a bit more explicit? i.e. https://lodash.com/docs/4.17.11#uniq

pages/pil/update/index.js
```diff
@@ -8,6 +8,8 @@ const procedures = require('../procedures');
 const exemptions = require('../exemptions');
 const training = require('../training');
 
+const flattenDeep = (arr) => { return arr.reduce((acc, val) => Array.isArray(val) ? acc.concat(flattenDeep(val)) : acc.concat(val), []); };
```
joefitter: https://lodash.com/docs/4.17.11#flattenDeep

pages/pil/update/index.js
lvnhmd: done

pages/pil/update/index.js
lvnhmd: done

PR #420, opened 2019-06-21
Amend PPL Holder

joefitter:
* added pages for PPL holder update journey
* added task content
* bumped projects and components

### DIFF
--- a/pages/place/routers/amend.js
+++ b/pages/place/routers/amend.js
@@ -1,6 +1,5 @@
 const form = require('../../common/routers/form');
 const { getSchemaWithNacwos, schema } = require('../schema');
-const { getNacwoById } = require('../../common/helpers');
 
 module.exports = settings => form(Object.assign({
   model: 'place',
@@ -9,15 +8,6 @@ module.exports = settings => form(Object.assign({
     getSchemaWithNacwos(req, settings.schema || schema)
       .then(mappedSchema => {
         req.form.schema = mappedSchema;
-      })
-      .then(() => next())
-      .catch(next);
-  },
-  locals: (req, res, next) => {
-    res.locals.static.schema = req.form.schema;
-    getNacwoById(req, req.form.values.nacwo)
-      .then(nacwo => {
-        res.locals.model.nacwo = nacwo;
       })
       .then(() => next())
       .catch(next);
--- a/pages/task/read/routers/view.js
+++ b/pages/task/read/routers/view.js
@@ -88,15 +88,6 @@ module.exports = () => {
         .then(() => next())
         .catch(next);
     }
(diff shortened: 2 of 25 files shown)


pages/place/routers/amend.js
```diff
@@ -13,15 +12,6 @@ module.exports = settings => form(Object.assign({
...
   },
-  locals: (req, res, next) => {
-    res.locals.static.schema = req.form.schema;
-    getNacwoById(req, req.form.values.nacwo)
-      .then(nacwo => {
-        res.locals.model.nacwo = nacwo;
-      })
-      .then(() => next())
-      .catch(next);
-  },
```
lennym: This seems weirdly out of scope. What's occurring?

pages/task/read/routers/view.js
```diff
@@ -88,6 +88,15 @@ module.exports = () => {
...
     }
+    if (req.task.data.model === 'project' && req.task.data.action === 'update') {
+      const licenceHolderId = get(req.task, 'data.data.licenceHolderId');
+      return req.api(`/establishment/${req.establishmentId}/profile/${licenceHolderId}`)
+        .then(({ json: { data } }) => {
+          set(req.task, 'data.data.licenceHolder', data);
+        })
+        .then(() => next())
+        .catch(next);
+    }
```
lennym: I'd be more inclined to do this in a workflow decorator and avoid the extra request round trip.

pages/place/routers/amend.js
joefitter: the fieldset was expecting an object for nacwo, I removed this so we always deal with the id except when rendering

pages/task/read/routers/view.js
joefitter: fair.

pages/task/read/views/project.jsx
```diff
@@ -17,45 +22,112 @@ const Project = ({ task, project, establishment, children, schema, formFields })
...
+        task.data.action === 'update' && (
+          <StickyNavAnchor id="licence-holder">
+            <h2><Snippet>sticky-nav.licence-holder</Snippet></h2>
+            <Diff values={task.data.data} model={project} schema={pick(projectSchema, 'licenceHolder')} formatters={formatters} />
+          </StickyNavAnchor>
+        )
+      }
+
+      {
+        task.data.action === 'update' && (
```
lennym: We have the same condition twice in a row here. I'm guessing this is because the Sticky Nav needs direct children and can't handle a `<Fragment />`?

joefitter: yes. it massively sucks and I fully intend to fix it, just not right now

PR #421, opened 2019-06-21
PEL authorisations

wheelsandcogs:
Allow the user to amend PEL authorisations.

### DIFF
--- a/pages/establishment/update/index.js
+++ b/pages/establishment/update/index.js
@@ -1,10 +1,35 @@
-const { omit } = require('lodash');
+const { pick, uniq } = require('lodash');
 const { page } = require('@asl/service/ui');
 const { form } = require('../../common/routers');
 const schema = require('./schema');
 const confirm = require('./routers/confirm');
 const success = require('../../common/routers/success');
 const { groupFlags, ungroupFlags } = require('../../establishment/formatters/flag-grouping');
+
+const fieldsToAuthorisations = params => {
+  return Object.keys(params).reduce((authorisations, fieldName) => {
+    if (params[fieldName] && /^authorisation-/.test(fieldName)) {
+      const id = fieldName.slice(-36); // get the uuid
+      let authorisation = authorisations.find(a => a.id === id);
+
+      if (!authorisation) {
+        authorisation = { id };
+        authorisations.push(authorisation);
+      }
+
+      authorisation.type = /killing/.test(fieldName) ? 'killing' : 'rehomes';
+
+      if (/method/.test(fieldName)) {
+        authorisation.method = params[fieldName];
+      }
+
+      if (/description/.test(fieldName)) {
+        authorisation.description = params[fieldName];
+      }
+    }
+    return authorisations;
+  }, []);
+};
 
 module.exports = settings => {
   const app = page({
@@ -16,6 +41,10 @@ module.exports = settings => {
   app.use((req, res, next) => {
     req.breadcrumb('establishment.update');
     req.model = groupFlags(req.establishment);
+
+    req.model.authorisationTypes = uniq(
+      req.model.authorisations.reduce((acc, authorisation) => acc.concat(authorisation.type), [])
+    );
 
     next();
   });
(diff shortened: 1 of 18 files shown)


pages/establishment/update/index.js
```diff
@@ -1,11 +1,36 @@
...
 const { form } = require('../../common/routers');
 const schema = require('./schema');
 const confirm = require('./routers/confirm');
 const success = require('../../common/routers/success');
 const { groupFlags, ungroupFlags } = require('../../establishment/formatters/flag-grouping');
 
+const fieldsToAuthorisations = params => {
+  return Object.keys(params).reduce((authorisations, fieldName) => {
+    if (params[fieldName] && /^authorisation-/.test(fieldName)) {
+      const id = fieldName.slice(-36); // get the uuid
```
joefitter: 🤢

pages/establishment/update/index.js
```diff
@@ -17,6 +42,10 @@ module.exports = settings => {
     req.breadcrumb('establishment.update');
     req.model = groupFlags(req.establishment);
 
+    req.model.authorisationTypes = uniq(
+      req.model.authorisations.reduce((acc, authorisation) => acc.concat(authorisation.type), [])
```
joefitter: .map

joefitter: `req.model.authorisations.map(authorisation => authorisation.type)`

pages/establishment/update/index.js
wheelsandcogs: it sucks but I can't think of another way - any suggestions?

pages/establishment/update/index.js
wheelsandcogs: yeah not sure what I was doing here!

pages/establishment/update/index.js
lennym: I would at the very least validate that the substring actually _is_ a UUID, to avoid the possibility of me POSTing fields named `authorisation-drop table projects;` or similar. At the moment you're just trusting that the last 36 characters of any field with the right prefix are valid.

lennym: Or match against `/^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/` in the regex.

lennym:
```js
const re = /^authorisation-([0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})$/;
const matched = fieldName.match(re);
if (matched) {
  const id = matched[1]; // I think, double check this
  /// etc
}
```

wheelsandcogs: There's other shizzle in the middle so i'll just confirm it's a uuid before using the field.

joefitter: Lenny's solution is a _lot_ better than the .slice(-36) - could we use a variation of this?

wheelsandcogs: There's now a check below for uuid but ok

pages/establishment/update/index.js
```diff
@@ -1,11 +1,43 @@
...
+const fieldsToAuthorisations = params => {
+  return Object.keys(params).reduce((authorisations, fieldName) => {
+    if (params[fieldName] && /^authorisation-/.test(fieldName)) {
+      const matched = fieldName.match(/^authorisation-(killing|rehomes)-(method|description)-([0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})$/);
+
+      if (!matched) {
+        return authorisations;
+      }
+
+      const [, authType, authInfo, id] = matched;
```
lennym: Does that work?

wheelsandcogs: yeah baby!

joefitter: #jode^10

PR #422, opened 2019-06-24
Connect species

lvnhmd:


pages/pil/exemptions/views/modules.jsx
```diff
@@ -13,18 +13,27 @@ import { Select } from '@ukhomeoffice/react-components';
 import { species } from '@asl/constants';
 
 import { normalise } from '../../../../lib/utils';
-
+const content = require('../content/modules');
 const SPECIES_REVEAL_TOTAL_COUNT = 10;
-const SPECIES_REVEAL_VISIBLE_COUNT = 1;
+const SPECIES_REVEAL_VISIBLE_COUNT = 0;
```
wheelsandcogs: Does this mean we don't show any species field by default?

pages/pil/exemptions/content/index.js
```diff
@@ -13,5 +13,11 @@ module.exports = {
     exempt: {
       required: 'You need to select yes or no.'
     }
+  },
+  action: {
+    repeat: {
+      add: 'Add another',
+      remove: 'Remove'
+    }
```
joefitter: can we bring this up a level so it can be shared by both exemptions and training?

pages/pil/exemptions/views/modules.jsx
```diff
@@ -1,60 +1,138 @@
...
+  label: content.fields.species.label
 };
 
+class RepeatedFieldset extends Component {
+
+  constructor(options) {
+    super(options);
+
+    this.state = {
+      items: this.props.model[`module-${this.props.type}-species`]
```
joefitter: this looks like it could be quite fragile - could we use `get`:
```js
this.state = {
  items: get(this.props, `model[module-${this.props.type}-species]`, [])
}
```

pages/pil/training/content/modules.js
```diff
@@ -18,5 +18,11 @@ module.exports = merge({}, content, {
     modules: {
       required: 'You need to select at least one module.'
     }
+  },
+  action: {
+    repeat: {
+      add: 'Add another',
+      remove: 'Remove'
+    }
```
joefitter: see above - this should be shared

pages/pil/training/routers/modules.js
```diff
@@ -31,6 +31,15 @@ module.exports = settings => {
...
       next();
+    },
+    process: (req, res, next) => {
+      req.form.values.modules.map(m => {
+        const specs = Object.values(pickBy(req.body, (value, key) => {
+          return startsWith(key, `module-${m}-species`);
+        })).filter(s => s !== '');
+        req.form.values[`module-${m}-species`] = specs;
+      });
+      next();
```
joefitter: could we extract this logic into a shared middleware?

pages/pil/training/views/modules.jsx
```diff
@@ -1,43 +1,138 @@
...
-                  name={`module-${op.value}-species`}
-                  label={<Snippet>{`fields.species.label`}</Snippet>}
-                  options={species}
-                />
-              </AddAnother>
-            </Inset>
-          ) : null
+          reveal: <Inset>
+            <ConnectedComponent type={op.value} />
+          </Inset>
```
joefitter: You said it!

All this repeated stuff should be in a shared module

pages/pil/exemptions/content/index.js
lvnhmd: done

pages/pil/exemptions/views/modules.jsx
lvnhmd: done

pages/pil/training/content/modules.js
lvnhmd: done

pages/pil/training/routers/modules.js
lvnhmd: done

pages/pil/training/views/modules.jsx
lvnhmd: done

lib/middleware/species.js
```diff
@@ -0,0 +1,17 @@
...
+    })).filter(s => s !== '');
+    req.form.values[`module-${m}-species`] = specs;
+  });
+
+  next();
+};
+
+module.exports = {
+  process
+};
```
joefitter: This seems like the logic is in the wrong place - we're processing the form, then manually adding species after using this middleware. Firstly we should let the form do the processing by adding the pseudo fields to the schema before processing (in configure), then we should reset the schema to default after processing (in locals) - this way we don't lose the things we get for free with the for router such as validation, module cleaning etc.

Another point is this is a middleware which is extremely dependent on the environment in which it is called - it would be better to either export a function, or perhaps include a function in the schema file which takes modules and exports schema extended with the new fields

lvnhmd: done

PR #429, opened 2019-06-26
Afternoon content amends.

pcuk1979:
Again, let me know if any of this doesn't work or breaks any tests.

### DIFF
--- a/pages/place/content/fields.js
+++ b/pages/place/content/fields.js
@@ -19,7 +19,7 @@ module.exports = {
   changesToRestrictions: {
     label: 'Changes to restrictions',
     conditionalReveal: {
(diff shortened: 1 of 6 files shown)


pages/place/content/fields.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
   changesToRestrictions: {
     label: 'Changes to restrictions',
     conditionalReveal: {
-      label: 'Do these restrictions need to be changed?',
+      label: 'Does this area\'s restrictions need to be changed?',
```
lennym: Should this be "Do this area's restrictions need to be changed?" since the restrictions are plural?

pcuk1979: I should have changed this now

PR #433, opened 2019-06-28
Fix PPL deadline bugs

lennym:
The deadline was being calculated base don initial submission, and not being reset if resubmitted.

The activity log was showing "With inspectorate" as the action for a deadline extension.

https://jira.digital.homeoffice.gov.uk/browse/AS-1609
https://jira.digital.homeoffice.gov.uk/browse/AS-1616


joefitter:
all your base don are belong to us

PR #434, opened 2019-06-28
Minor content changes

pcuk1979:
FROM AS-1548


lennym:
It feels like there's some potential for optimisation here...

PR #437, opened 2019-07-02
Display a message if user doesn't have dob or is under 18

wheelsandcogs:


pages/pil/create/index.js
```diff
@@ -1,11 +1,22 @@
...
   const app = page({
     ...settings,
     root: __dirname
   });
 
+  app.use((req, res, next) => {
+    const over18 = req.profile.dob ? differenceInYears(new Date(), new Date(req.profile.dob)) >= 18 : false;
+
+    if (!over18) {
+      throw new Error('All licence applicants must be over 18.');
```
wheelsandcogs: This should only happen if they somehow directly hit the route as there should be no "Apply for PIL" button available to them.

Taking that into account, is this error message adequate, or should we reproduce all the variants we have in the profile page (e.g. no dob set, dob set but under 18, own profile, not own profile, etc)?

joefitter: This message seems fine to me, especially as it is highly unlikely to ever be seen

PR #438, opened 2019-07-03
Clean up profile view for dashboard

wheelsandcogs:
Makes a few requested design changes to the profile that gets embedded in the user's dashboard.

### DIFF
--- a/pages/profile/read/views/profile.jsx
+++ b/pages/profile/read/views/profile.jsx
@@ -42,20 +42,13 @@ class Profile extends React.Component {
 
               {
                 activeProjects.map(project => (
-                  <Fragment key={project.id}>
+                  <div key={project.id} className="project">
                     <p>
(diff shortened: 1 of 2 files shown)


pages/profile/read/views/profile.jsx
```diff
@@ -102,7 +95,7 @@ class Profile extends React.Component {
                 isOwnProfile && (
                   <form method='POST' action={`/e/${estId}/projects/create`}>
                     <p className="control-panel">
-                      <Button className='govuk-button add-margin'>
+                      <Button className='govuk-button button-secondary add-margin'>
```
joefitter: not going to block this, as the code was already there, but if you use a `<Button />` you don't need to add the govuk-button class as this is added by default

wheelsandcogs: 👍 noted

PR #446, opened 2019-07-10
Changes to inline states

pcuk1979:
The default text for noDob, under18, and addDob have been changed.

I have removed the 'Can't apply for licence' heading.

Note that the logic here needs changing. The only time that 'None' would display in this component is if a user has a DoB in their account, is over 18, but has not yet applied for a PIL. In all other cases, 'None' should be omitted.

### DIFF
--- a/pages/profile/read/content/index.js
+++ b/pages/profile/read/content/index.js
@@ -41,12 +41,11 @@ module.exports = {
       incomplete: 'Application started.',
       notStarted: 'None'
     },
-    cantApply: 'Can\'t apply for licence',
     noDob: {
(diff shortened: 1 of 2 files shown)


lennym:
I think this is going to need some work to not break - this line needs removing also https://github.com/UKHomeOffice/asl-pages/blob/master/pages/profile/read/views/profile.jsx#L189

lennym:
I'm not sure I understand the changes required to the logic either. Is it that the "None" label should _only_ show in cases where a user _can_ apply for a PIL, and not in cases where there is some reason why they cannot?

PR #447, opened 2019-07-10
Discard project drafts and amendments

wheelsandcogs:


pages/project/delete/index.js
```diff
@@ -0,0 +1,23 @@
...
+    if (req.project.status === 'inactive') {
+      return req.api(`/establishment/${req.establishmentId}/project/${req.projectId}`, { method: 'DELETE' })
+        .then(() => res.redirect(`${req.buildRoute('project.list')}?status=inactive`))
+        .catch(next);
+    }
+
+    throw new Error('Active projects cannot be deleted.');
+  });
+
+  app.use('/amendment', (req, res, next) => {
```
lennym: You don't want `app.use` here because that will listen to all HTTP verbs. You'll want `app.post`.

lennym: Similar above.

lennym:
Do we want to display a "success" toaster post-delete?

wheelsandcogs:
They get redirected to the list of drafts which their deleted draft will be missing from, or in the case of discarding amendments, all the wording around amendments on the project landing page changes (e.g. the button will change from "Continue editing amendment" to "Amend licence").

It feels obvious enough to me, but maybe we want one for consistency?

lennym:
_I_ would like one. Given it's a one-line code change.

PR #455, opened 2019-07-15
dont pass granted flag for legacy ppls

joefitter:


pages/project-version/read/index.js
```diff
@@ -32,7 +32,9 @@ module.exports = settings => {
     res.locals.model = req.version;
     res.locals.static.project = req.project;
     res.locals.static.version = req.version.id;
-    res.locals.static.isGranted = req.project.status === 'active' && req.project.granted.id === req.version.id;
+    res.locals.static.isGranted = req.project.status === 'active' &&
+      req.project.granted.id === req.version.id &&
+      req.project.schemaVersion > 0;
```
lennym: Y THO?

Can you just add a comment to explain this pls, because on reading the code it looks super weird.

lennym:
Generally I'd probably prefer to check the version at the point at which the `isGranted` property is used, rather than bundling it in here, but also to some extent whatevs.

joefitter:
yeah that probably would make more sense, but there were about 10ish places and I was being lazy

PR #464, opened 2019-07-23
Fix model

lvnhmd:


pages/establishment/update/views/index.jsx
```diff
@@ -105,7 +105,7 @@ class Repeat extends Component {
 
               <Fieldset
                 {...this.props}
-                model={mapKeys(authorisations.filter(auth => auth.type !== undefined), (val, key) => `authorisation-${authorisation.type}-${key}-${authorisation.id}`)}
+                model={mapKeys(authorisations.filter(auth => auth.type !== undefined)[index], (val, key) => `authorisation-${authorisation.type}-${key}-${authorisation.id}`)}
```
lennym: Taking the same index after filtering the array is likely to break by returning an unexpected entry. I don't understand what the underlying issue is, but my best guess is that you want to retreive the entry at the index, and then cehck its type property.

lvnhmd: obsolete, it turned out to be another bug

pages/establishment/update/views/index.jsx
```diff
@@ -74,8 +74,7 @@ class Repeat extends Component {
         authorisations: authorisations.map((authorisation, i) => {
           if (i === index) {
             return {
-              ...authorisation,
-              ...model
+              ...authorisation
```
lennym: This isn't doing anything at all to the state. Either `i === index` and it returns the authorisation, or it doesn't and it returns... the authorisation.

There's no update.

lvnhmd: Removed onFieldsetChange - we do not need it.

PR #468, opened 2019-07-25
Content amends

pcuk1979:
All around revocation 'requests'

### DIFF
--- a/pages/task/content/status.js
+++ b/pages/task/content/status.js
@@ -23,7 +23,7 @@ module.exports = {
     hint: {
       application: 'The application will be sent to an inspector for assessment.',
       amendment: 'The amendment will be sent to an inspector for assessment.',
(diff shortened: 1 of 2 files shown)


pages/task/content/status.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
     hint: {
       application: 'The application will be sent to an inspector for assessment.',
       amendment: 'The amendment will be sent to an inspector for assessment.',
-      revocation: 'The revocation will be sent to an inspector for assessment.'
+      revocation: 'The revocation request will be sent to an inspector for assessment.'
```
lennym: Is that true?

lennym: Sorry, my bad, I was missing the context.

I thought this was on an initial submission of the request, but it isn't.

PR #470, opened 2019-07-26
Feature/project download bar

joefitter:


pages/project-version/docx/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+      .then(pack)
+      .then(buffer => {
+        res.attachment(`${values.title || 'Untitled project'}.docx`);
+        res.end(Buffer.from(buffer));
+      })
+      .catch(next);
+  });
+
+  return app;
+};
```
lennym: This probably shouldn't be a `.jsx` file

joefitter: no, it should not 👍

PR #474, opened 2019-07-29
If the latest version of the project is withdrawn, set it to draft

lvnhmd:


pages/project/read/views/index.jsx
```diff
@@ -92,6 +92,15 @@ const App = ({ model, establishment, url, content, openTask, canAmend, canDelete
...
 
+        {
+          !model.granted && model.withdrawn &&
+            <form method="POST">
+              <button className="govuk-button">
+                <span><Snippet>fields.draft.view</Snippet></span>
+              </button>
+            </form>
+        }
+
```
lennym: This appears that viewing the withdrawn version will _always_ create a new draft. It's likely we'll want separate "Continue editing" and "Just look at it" calls to action.

lennym: Also, the actual editing will need to be permission based as well as just dependent on the state. It shouldn't create a fork for all users. 

I'd look at moving this inside the `canAmend` condition further down.

lvnhmd: > This appears that viewing the withdrawn version will _always_ create a new draft. It's likely we'll want separate "Continue editing" and "Just look at it" calls to action.

@lennym , I think this is _different_ from the original requirement which was to show a button to open draft in *edit* mode - I remember asking should I fork it at the moment of opening and we agreed that I should , let's discuss

lennym: That requirement really means that "people who _can_ edit the project should be able to do that". You are displaying this button to everyone, including people who do not have permission to edit the project.

Those people should still be able to _see_ the version, but not be able to edit it.

lvnhmd: Done

wheelsandcogs: I believe this is correct now, gonna demo every variation to someone

PR #483, opened 2019-08-02
Bugfix/show pils

joefitter:


pages/pil/read/content/index.js
```diff
@@ -12,6 +12,7 @@ module.exports = merge({}, baseContent, {
     conditions: { label: 'Conditions' },
     procedures: { label: 'Procedures' }
   },
+  cantUpdate: '**You cant amend this PIL as it is held at another establishment**',
```
lennym:
```suggestion
  cantUpdate: '**You can\'t amend this PIL as it is held at another establishment**',
```

lennym:
```suggestion
  cantUpdate: '**You cannot amend this PIL as it is held at another establishment**',
```

lennym: One of those ☝️

joefitter: OOPS

pages/pil/update/index.js
```diff
@@ -29,7 +29,13 @@ module.exports = settings => {
       establishment: req.establishment.id
```
lennym:
```suggestion
      establishment: req.pil.establishmentId
```

pages/pil/update/index.js
```diff
@@ -29,7 +29,13 @@ module.exports = settings => {
...
     };
     req.user.can('pil.update', params)
-      .then(can => can ? next() : next(new Error('Unauthorised')))
+      .then(can => {
+        const authorised = can && req.pil.establishmentId === req.establishment.id;
```
lennym: I'd probably expect this check to be incorporated into the permissions API, which to my understanding should till check that a user is either the PIL holder or an admin at the holding establishment.

I believe this can be done with the `params` change above.

pages/profile/read/content/index.js
```diff
@@ -36,6 +36,7 @@ module.exports = merge({}, baseContent, {
     title: 'Personal licence',
     noPil: 'None',
     incompletePil: 'Application started.',
+    incompleteOtherEst: 'Application started at another establishment',
```
lennym: If "Application started." has a full stop then this probably should too.
```suggestion
    incompleteOtherEst: 'Application started at another establishment.',
```

pages/profile/read/views/profile.jsx
```diff
@@ -199,7 +200,11 @@ class Profile extends React.Component {
                 )
               }
               {
-                (isOwnProfile || allowedActions.includes('pil.create')) && !pilActive && over18 && (
+                (isOwnProfile || allowedActions.includes('pil.create')) &&
+                  !pilActive &&
+                  over18 &&
+                  (correctEstablishment || !pilIncomplete) &&
```
lennym: This is getting pretty burly now. Perhaps pull it out into a `const canApply`?

joefitter: I thought you might have :cow2: with that one

PR #486, opened 2019-08-02
Don't check for "ready" declaration for amendments

wheelsandcogs:


pages/task/read/views/project.jsx
```diff
@@ -27,9 +27,16 @@ const completeAndCorrect = task => {
 // declarations can be 'Yes', 'No', or 'Not yet'
 const declarationConfirmed = declaration => declaration.toLowerCase() === 'yes';
```
lennym:
```suggestion
const declarationConfirmed = declaration => declaration && declaration.toLowerCase() === 'yes';
```

Since that's the actual root cause of your bug.

PR #488, opened 2019-08-02
Don't show the deadline if the task is an amendment

wheelsandcogs:
Also, don't show the "ready" declaration for amendments.

### DIFF
--- a/pages/task/read/views/project.jsx
+++ b/pages/task/read/views/project.jsx
@@ -11,14 +11,14 @@ import ReviewFields from '@asl/projects/client/components/review-fields';
 import { fields } from '../../../project/update-licence-holder/schema/experience-fields';
 import { schema as projectSchema } from '../../../project/schema';
 
-const completeAndCorrect = (task, project) => {
+const completeAndCorrect = (task, isAmendment) => {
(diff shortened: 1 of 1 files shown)


pages/task/read/views/project.jsx
```diff
@@ -144,7 +146,7 @@ const Project = ({ task, project, establishment, children, schema, formFields })
       }
 
       {
-        completeAndCorrect(task, project) &&
+        !isAmendment && completeAndCorrect(task, isAmendment) &&
```
lennym: Do we use `completeAndCorrect` anywhere _except_ immediately after checking it's not an amendment?

If not then the second argument might be redundant.

wheelsandcogs: Yeah I was debating this, as it's currently only called for new applications. 

But if we called `allDeclarationsConfirmed` somewhere else in future without realising we need to deal with `isAmendment` then we'll end up with a bug, so I thought it best to make it obvious while it's fresh in my mind.

lennym: That actually seems pretty reasonable. 

👍

PR #490, opened 2019-08-05
Render awerb review dates as markdown

wheelsandcogs:


pages/task/read/views/project.jsx
```diff
@@ -117,7 +118,8 @@ const Project = ({ task, project, establishment, children, schema, formFields })
                     declarationConfirmed(declarations.awerb) &&
                       <Fragment>
                         <p>
-                          <Snippet>declarations.awerb.review-date</Snippet> {declarations['awerb-review-date']}
+                          <Snippet>declarations.awerb.review-date</Snippet>
+                          <ReactMarkdown>{declarations['awerb-review-date']}</ReactMarkdown>
```
joefitter: @wheelsandcogs I think Snippet renders as markdown by default - you probably only need one snippet and pass the review data in as a variable

lennym: I think that's a label and then a value, so this seems correct to me.

However, rendering `Markdown` inside a `<p>` is likely to result in some mild weirdness because the markdown will render its own `<p>` tags (which is the point).

wheelsandcogs: sorted.

PR #492, opened 2019-08-06
Handle workflow connection errors

joefitter:
If workflow is unavailable the api now still responds but with a connectionError if trying to load all tasks, and a single task with action: 'failed' when fetching open tasks for a model.

Handle these cases by showing an alternate component for the tasklist and different message when trying to edit a model

### DIFF
--- a/pages/common/components/in-progress-warning.jsx
+++ b/pages/common/components/in-progress-warning.jsx
@@ -3,7 +3,15 @@ import { Link } from '@asl/components';
 
 export default ({ task }) => {
   return <Fragment>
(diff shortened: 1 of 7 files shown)


lennym:
There's a couple of cases of pretty big ternary operators being used here where the more appropriate pattern feels like it's an early return of an error state.

i.e. the old node-callback pattern of:

```
if (err) {
  return handle(err);
}
// happy path
```

I think particularly in the "in progress warning" where it feels like there should definitely be the possibility for other states, and so a ternary doesn't make real sense.

PR #496, opened 2019-08-08
Hide deadline extend button on closed tasks

wheelsandcogs:
Waiting on https://github.com/UKHomeOffice/asl-workflow/pull/130

### DIFF
--- a/pages/task/read/views/deadline.jsx
+++ b/pages/task/read/views/deadline.jsx
@@ -5,14 +5,6 @@ import { dateFormat } from '../../../../constants';
 import { formatDate } from '../../../../lib/utils';
 import moment from 'moment-business-time';
 
-const withAsru = task => [
-  'with-inspectorate',
-  'referred-to-inspector',
-  'with-licensing',
-  'inspector-recommended',
-  'inspector-rejected'
-].includes(task.status);
-
 class Deadline extends Component {
   isExtended() {
     return !!this.props.task.data.extended;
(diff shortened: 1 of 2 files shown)


pages/task/read/views/deadline.jsx
```diff
@@ -5,6 +5,14 @@ import { dateFormat } from '../../../../constants';
 import { formatDate } from '../../../../lib/utils';
 import moment from 'moment-business-time';
 
+const withAsru = task => [
+  'with-inspectorate',
+  'referred-to-inspector',
+  'with-licensing',
+  'inspector-recommended',
+  'inspector-rejected'
+].includes(task.status);
```
lennym: This feels fragile to me, in that it will potentially break if we modify workflow with new states. 

We're specifically interested in tasks not being closed (i.e. that they're not in one of the end-state statuses) but I think it's probably better to expose that out of workflow as a Boolean flag on a task rather than leaking the logic all the way to the browser.

lennym: As an aside, I think there's generally too much business logic going on client-side in this component - i.e. complete and correct-ness, deadline calculation etc - that could and probably should be done server side and lighten the payload a little.

Obviously out of scope for this ticket though.

wheelsandcogs: Agree that it's fragile, I'll add a flag.

wheelsandcogs: https://github.com/UKHomeOffice/asl-workflow/pull/130

PR #501, opened 2019-08-13
Simplified project init - now only take initial state

joefitter:


joefitter:
Waiting on https://github.com/UKHomeOffice/asl-projects/pull/343

PR #515, opened 2019-08-28
Revoke a PPL

wheelsandcogs:
~Waiting on https://github.com/UKHomeOffice/asl-public-api/pull/173~

### DIFF
--- a/pages/project/revoke/index.js
+++ b/pages/project/revoke/index.js
@@ -11,7 +11,9 @@ module.exports = () => {
 
   app.use((req, res, next) => {
     req.breadcrumb('project.revoke');
-    req.model = req.project;
+    req.model = {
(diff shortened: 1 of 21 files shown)


pages/project/revoke/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+module.exports = () => {
+  const app = page({
+    root: __dirname,
+    paths: ['/confirm', '/success']
+  });
+
+  app.use((req, res, next) => {
+    req.breadcrumb('project.revoke');
+    req.model = req.project;
```
joefitter: could we use an empty object here? the `model` in a form only relates to the form values, it also uses the model id for session storage and cleanup - I'd go with

```
req.model = {
  id: `${req.project.id}-revoke`
};
```

PR #525, opened 2019-09-10
PIL PDF view

wheelsandcogs:
Still to do:
 * ~Add download link to PIL read view (currently needs manual `/pdf` adding to the url)~
 * ~Fix page breaks~
 * ~Fix markdown rendering of the legal preamble~


package.json
```diff
@@ -8,11 +8,13 @@
...
-    "build:css": "npm-sass ./pages/common/assets/sass/style.scss > ./pages/common/dist/css/app.css",
-    "prebuild:css:pdf": "mkdir -p ./pages/project-version/pdf/css",
-    "build:css:pdf": "node-sass ./pages/project-version/pdf/sass -o ./pages/project-version/pdf/css",
+    "prebuild": "mkdir -p ./pages/common/dist/css/pdf",
+    "build": "npm run build:css",
+    "build:css": "npm run build:css:app && npm run build:css:pdf && npm run build:css:pdf:project",
+    "build:css:app": "npm-sass ./pages/common/assets/sass/style.scss > ./pages/common/dist/css/app.css",
+    "build:css:pdf": "node-sass ./pages/common/assets/sass/pdf -o ./pages/common/dist/css/pdf",
+    "prebuild:css:pdf:project": "mkdir -p ./pages/project-version/pdf/css",
+    "build:css:pdf:project": "node-sass ./pages/project-version/pdf/sass -o ./pages/project-version/pdf/css",
```
wheelsandcogs: Once this is merged I'll do another ticket to make projects use the new generic code, and remove the two `pdf:project` lines.

pages/common/pdf/client/reducers/index.js
```diff
@@ -0,0 +1,12 @@
...
+import licenceHolder from './profile';
+import establishment from './establishment';
+
+const rootReducer = combineReducers({
+  pil,
+  licenceHolder,
+  establishment
+});
+
+export default rootReducer;
```
wheelsandcogs: These reducers don't really do anything, but if they don't exist, `createStore()` starts throwing warnings in the log? I've kept them for now to get it working but I'm expecting we'll remove / replace them.

pages/pil/pdf/views/pil.jsx
```diff
@@ -0,0 +1,116 @@
...
+
+  return (
+    <Fragment>
+      <div className="logo"></div>
+
+      <h3 className="licence">PERSONAL LICENCE</h3>
+
+      <h1 className="licence-holder">{`${licenceHolder.firstName} ${licenceHolder.lastName}`}</h1>
+
+      {/* <ReactMarkdown>{content.legalPreamble}</ReactMarkdown> */}
```
wheelsandcogs: If I include the text below via `ReactMarkdown` it seems to cause the styling to break, and I'm not sure why.

pages/common/pdf/client/reducers/index.js
joefitter: the app needs state, which it gets from the redux store, which this rootReducer creates

joefitter: however, as it's all static, you don't need 3 reducers to do one thing, just have one and call it static or something?

pages/common/pdf/client/store.js
```diff
@@ -0,0 +1,14 @@
...
+import thunk from 'redux-thunk';
+import logger from 'redux-logger';
+
+const middlewares = [thunk, logger];
+
+export default function configureStore(initialState = {}) {
+  return createStore(
+    reducer,
+    initialState,
+    applyMiddleware(...middlewares)
```
lennym: I'm _fairly_ sure we don't need these middlewares.

pages/common/pdf/client/reducers/index.js
joefitter: all this is doing is creating a static state object that looks like this: 
```js
{
  pil: {},
  establishment: {},
  ...
}
```

package.json
joefitter: it would be better to sort them out before merging IMO - there's gunna be a _lot_ of repeated stuff otherwise

pages/common/pdf/client/reducers/index.js
joefitter: actually, thinking about this, it doesn't need redux at all, all of the store stuff can go, this is 100% new code so doesn't have to follow the same pattern as our complex application - you could just pass in props to the <App /> on init

pages/pil/pdf/content/standard-conditions.js
```diff
@@ -0,0 +1,92 @@
...
+  appropriate. This record shall be retained for at least five years and shall, on request, be submitted to the
+  Secretary of State or made available to an Inspector.`,
+
+  `The licence holder must give any necessary assistance to inspectors carrying out visits by virtue of
+  section 18(2A)(b) ; and to experts of the European Commission carrying out duties under Article 35 of
+  the Animals Directive.`,
+
+  `The licence remains the property of the Secretary of State, and shall be surrendered to him on
+  request.`
+];
```
lennym: What's the logic behind this being an array and not just a big block of markdown?

pages/pil/read/views/index.jsx
```diff
@@ -19,7 +19,8 @@ const PIL = ({
   canUpdate,
   allowedActions,
   openTask,
-  correctEstablishment
+  correctEstablishment,
```
lennym: 👀 

As opposed to `incorrectEstablishment`?

My guess is that the establishment that holds the PIL appears on the pdf version? Not necessarily the establishment that is requesting the document?

package.json
wheelsandcogs: This is resolved in another open pages PR that migrates the project pdfs to use the generic code.

pages/pil/pdf/content/standard-conditions.js
wheelsandcogs: I was copying how @joefitter did his conditions here: https://github.com/UKHomeOffice/asl-projects/blob/master/client/constants/standard-conditions.js so that I could re-use the purple inset styling.

pages/pil/read/views/index.jsx
wheelsandcogs: Not my code, I just added `currentPath`.

package.json
wheelsandcogs: https://github.com/UKHomeOffice/asl-pages/pull/528/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

pages/pil/pdf/content/standard-conditions.js
lennym: 👍

pages/pil/read/views/index.jsx
lennym: Yeah, it just made me aware of the issue with potentially leaking establishment details.

pages/pil/pdf/views/pil.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+      <h3 className="licence">PERSONAL LICENCE</h3>
+
+      <h1 className="licence-holder">{`${licenceHolder.firstName} ${licenceHolder.lastName}`}</h1>
+
+      <ReactMarkdown className="legal-preamble">{content.legalPreamble}</ReactMarkdown>
+
+      <section className="primary-establishment break">
+        <h2>Primary establishment</h2>
+        <ReactMarkdown>{establishment.address}</ReactMarkdown>
```
lennym: This might need to go.

If it doesn't then it's super weird to only have the address and not the actual name of the establishment.

lennym: Sounds like we need to pass in the `correctEstablishment` prop here too and hide the primary establishment details if it's not the one requesting the document.

wheelsandcogs: Yeah we want the establishment name in there as well, I think I missed it cause "University of Croydon" is the address in our seeds.

PR #530, opened 2019-09-16
Bugfix/inspector revoking ppl

joefitter:
* show task info if inspector submitted revoke
* don't show revoke licence button on non active project

### DIFF
--- a/pages/project/read/index.js
+++ b/pages/project/read/index.js
@@ -30,26 +30,20 @@ module.exports = settings => {
       licenceHolderId: req.project.licenceHolderId,
       establishment: req.establishment.id
     };
-    Promise.resolve()
-      .then(() => req.user.can('project.update', params))
-      .then(canUpdate => {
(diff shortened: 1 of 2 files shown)


pages/project/read/index.js
```diff
@@ -30,26 +30,20 @@ module.exports = settings => {
       licenceHolderId: req.project.licenceHolderId,
       establishment: req.establishment.id
     };
-    Promise.resolve()
-      .then(() => req.user.can('project.update', params))
-      .then(canUpdate => {
+    Promise.all([
+      req.user.can('project.update', params),
+      req.user.can('project.revoke', params)
+    ])
```
lennym: This is because ASRU can't update but they _can_ revoke, right?

joefitter: yeah, we were only calling permissions service for revoke if the task was `editable` before, which didn't make sense for asru users

PR #532, opened 2019-09-16
Update change licence holder task

joefitter:
Added middlware for updating the licence holder task when editable by applicant

### DIFF
--- a/pages/project/update-licence-holder/routers/confirm.js
+++ b/pages/project/update-licence-holder/routers/confirm.js
@@ -1,11 +1,29 @@
 const { Router } = require('express');
-const { get, omit } = require('lodash');
+const { get, omit, merge } = require('lodash');
 const form = require('../../../common/routers/form');
 const schema = require('../schema/declaration');
 const experienceFields = require('../schema/experience-fields');
+const { updateDataFromTask, redirectToTaskIfOpen } = require('../../../common/middleware');
+
+const sendData = (req, params = {}) => {
+  const values = get(req.session, `form.${req.model.id}.values`);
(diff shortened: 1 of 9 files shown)


pages/project/update-licence-holder/routers/confirm.js
```diff
@@ -1,19 +1,38 @@
...
+  app.post('/', updateDataFromTask(sendData));
+
   app.use(form({
     schema,
     locals(req, res, next) {
       res.locals.static.fields = experienceFields.fields;
       res.locals.static.project = req.project;
       res.locals.static.values = get(req.session, `form.${req.model.id}.values`);
-      req.api(`/establishment/${req.establishmentId}/profiles/${res.locals.static.values.licenceHolder}`)
+      console.log(res.locals.static.values.licenceHolderId);
```
lennym:
```suggestion
```

lennym:
Is changing the name of the property from `licenceHolder` to `licenceHolderId` going to cause any backwards compatibility issues with existing tasks?

Is the reason for the renaming technically driven, or just a matter of making the code clearer?

joefitter:
the property is licenceHolderId in the task, the change is contained within the UI. A bit of both, it was bit odd we were calling it `licenceHolder` when the property we are updating is licenceHolderId - this removes the need for mapping the prop

lennym:
I think that makes sense. I see [the mapping](https://github.com/UKHomeOffice/asl-pages/pull/532/files#diff-b157f90c1d1ad73b8c938c2cbbe8a98bL31-L33) now.

PR #533, opened 2019-09-16
Establishment PDF view

wheelsandcogs:


pages/establishment/pdf/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+import Header from '../../common/views/pdf/header';
+import Footer from '../../common/views/pdf/footer';
+import content from './content';
+
+const namedPeople = establishment => {
+  const profiles = {};
+
+  ['pelh', 'holc', 'nprc', 'nio', 'ntco', 'nvs', 'nacwo'].forEach(roleType => {
+    profiles[roleType] = establishment.roles.filter(role => role.type === roleType).map(role => role.profile);
+  });
```
lennym: https://lodash.com/docs/4.17.15#groupBy

pages/establishment/pdf/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+  return profiles;
+};
+
+const groupBy = key => array =>
+  array.reduce((objectsByKeyValue, obj) => {
+    const value = obj[key];
+    objectsByKeyValue[value] = (objectsByKeyValue[value] || []).concat(obj);
+    return objectsByKeyValue;
+  }, {});
```
lennym: Also https://lodash.com/docs/4.17.15#groupBy

wheelsandcogs: heh :D

pages/common/views/pdf/header.jsx
```diff
@@ -17,7 +17,7 @@ const Header = ({ model, licenceType, nonce }) => (
 
         <Fragment> | Granted: <strong>{format(model.issueDate)}</strong></Fragment>
 
-        { licenceType === 'pil' &&
+        { ['pil', 'pel'].includes(licenceType) &&
           <Fragment> | Last amended: <strong>{format(model.updatedAt)}</strong></Fragment>
```
lennym: This will likely need to be updated somewhere to reflect the fact that roles, authorisations and places are also part of the PEL, and so just taking the `updated_at` from the establishments table is not sufficient to provide a "Last amended" on the pEL itself.

Probably do that in the resolver as another ticket though.

pages/establishment/pdf/index.jsx
wheelsandcogs: Only problem with `groupBy` here is that it won't populate role types that aren't assigned, so I'll have to check for their existence when I use em.

pages/establishment/pdf/views/establishment.jsx
```diff
@@ -0,0 +1,191 @@
...
+        {
+          ['procedure', 'breeding', 'supplying'].filter(auth => establishment[auth]).map(auth =>
+            <li key={auth}>{content.licenced[auth]}</li>
+          )
+        }
+      </ul>
+
+      <section className="licence-holder">
+        <h2>Establishment licence holder</h2>
+        <p><strong>{`${licenceHolder.firstName} ${licenceHolder.lastName}`}</strong></p>
```
lennym: The PEL holder only exists for some establishments. Others have no PEL holder and an NPRC instead.

Need to ensure this works ok for those establishments as well.

pages/establishment/pdf/views/establishment.jsx
```diff
@@ -0,0 +1,191 @@
...
+                        <th>Name</th>
+                        <th>Suitability</th>
+                        <th>Holding</th>
+                        <th>NACWO</th>
+                        <th>Restrictions</th>
+                      </tr>
+                    </thead>
+                    <tbody>
+                      {
+                        places.map(place => (
```
lennym: There's a desire that these are secondarily sorted by `area` and `name` - the API will give them back in database order.

wheelsandcogs: I'm adding a sort=site param when I fetch them, do we support secondary sort orders in the api? Better adding it there if not.

pages/establishment/pdf/index.jsx
lennym: Is that ok? It's possibly desirable that some roles - specifically NPRC - don't appear if there are none.

As an aside, `holc` probably shouldn't be in this list as it's not a licensed role.

pages/establishment/pdf/views/establishment.jsx
lennym: Yeah, this desire was reported as part of the digital view, so I'm inclined to agree that the API should probably apply those secondary sorts. 

But, a) that's not been done yet, and b) we'd still probably need to make sure that the sort order was preserved once it went through the `groupBy`

pages/establishment/pdf/views/establishment.jsx
```diff
@@ -0,0 +1,196 @@
...
+    if (first) {
+      first = false;
+      return '';
+    }
+    return ' (continued)';
+  };
+};
+
+const Establishment = ({ establishment, content }) => {
+  const licenceHolders = establishment.namedPeople.pelh || establishment.namedPeople.nprc || [];
```
lennym: Based on what you told me about the reasons for not using `groupBy` in the `namedPeople` function then `establishment.namedPeople.pelh` will always be truthy since it will be `[]` even if no PEL holder roles exist.

lennym:
```suggestion
  const licenceHolders = [].concat(establishment.namedPeople.pelh).concat(establishment.namedPeople.nprc);
```

lennym: That said - I'm not sure it's legit to describe an NPRC as a licence holder.

lennym: My instinct would be to only ever call explicit PEL holders "licence holders", and just display "None" if there are none.

We have the section for NPRCs under named people already so we're covered there.

lennym: So in fact...
```suggestion
  const licenceHolders = establishment.namedPeople.pelh;
```

pages/establishment/pdf/views/establishment.jsx
```diff
@@ -0,0 +1,196 @@
...
+        <p>{ establishment.namedPeople.ntco.map(getName).join(', ') || 'None.' }</p>
+
+        <h4>Named Veternary Surgeons</h4>
+        <p>{ establishment.namedPeople.nvs.map(getName).join(', ') || 'None.' }</p>
+
+        <h4>Named Animal Care and Welfare Officers</h4>
+        <p>{ establishment.namedPeople.nacwo.map(getName).join(', ') || 'None.' }</p>
+      </section>
+
+      {/* page break */}
```
lennym: Is this _actually_ creating a page break, or purely indicative?

pages/establishment/pdf/index.jsx
wheelsandcogs: Hold doesn’t get used but I’ll remove it

pages/establishment/pdf/views/establishment.jsx
wheelsandcogs: Just a comment

pages/establishment/pdf/index.jsx
lennym: I see HOLC isn't rendered anyway. We should remove it from this list.

pages/establishment/pdf/views/establishment.jsx
lennym: I've created a ticket for doing the secondary sort in the API.

pages/establishment/pdf/views/establishment.jsx
wheelsandcogs: Class ‘.break’ adds the page break after the element it’s applied to

pages/establishment/pdf/index.jsx
wheelsandcogs: Please can you add a suggestion with it removed and I’ll commit it via phone 🙂

pages/establishment/pdf/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+import groupBy from 'lodash/groupBy';
+import Body from './views';
+import Header from '../../common/views/pdf/header';
+import Footer from '../../common/views/pdf/footer';
+import content from './content';
+
+const namedPeople = establishment => {
+  const profiles = {};
+
+  ['pelh', 'holc', 'nprc', 'nio', 'ntco', 'nvs', 'nacwo'].forEach(roleType => {
```
joefitter:
```suggestion
  ['pelh', 'nprc', 'nio', 'ntco', 'nvs', 'nacwo'].forEach(roleType => {
```

pages/common/views/pdf/header.jsx
joefitter: was a ticket added for this?

PR #539, opened 2019-09-20
Change the updatedAt column name in the tasklist

wheelsandcogs:
...depending on who is viewing and which tab they are on.

### DIFF
--- a/pages/task/list/router.js
+++ b/pages/task/list/router.js
@@ -27,6 +27,7 @@ module.exports = ({
     next();
   },
   locals: (req, res, next) => {
+    const progress = req.query.progress || tabs(req.user.profile)[0];
     const firstName = get(req, 'user.profile.firstName');
     const lastName = get(req, 'user.profile.lastName');
     res.locals.static.profileName = `${firstName} ${lastName}`;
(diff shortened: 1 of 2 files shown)


pages/task/list/router.js
```diff
@@ -27,13 +27,21 @@ module.exports = ({
...
     const lastName = get(req, 'user.profile.lastName');
     res.locals.static.profileName = `${firstName} ${lastName}`;
     res.locals.static.progress = req.datatable.progress;
     res.locals.datatable.progress = req.datatable.progress;
 
     res.locals.static.tabs = tabs(req.user.profile);
+
+    if (progress === 'completed') {
+      res.locals.static.content.fields.updatedAt.label = 'Completed';
+    } else if (progress === 'inProgress' || (progress === 'outstanding' && !req.user.profile.asruUser)) {
```
lennym: This is assuming a perfect flow through the system, but falls over if a task follows an unhappy path.

If a task is returned by a licensing officer to an inspector then it will appear in their `outstanding` tab, but the date will show when it was referred, not when it was submitted.

It feels to me like the label should always be either "Completed" or "Last changed" if the field that is being shown in the `updated_at`.

wheelsandcogs: Good point. I'll make it Last changed and you can explain to design ;)

lennym: I'm fine with that.

pages/task/list/router.js
```diff
@@ -34,6 +34,13 @@ module.exports = ({
     res.locals.datatable.progress = req.datatable.progress;
 
     res.locals.static.tabs = tabs(req.user.profile);
+
+    const progress = req.query.progress || tabs(req.user.profile)[0];
```
lennym:
```suggestion
```

pages/task/list/router.js
```diff
@@ -34,6 +34,13 @@ module.exports = ({
...
+
+    if (progress === 'completed') {
```
lennym:
```suggestion
    if (req.query.progress === 'completed') {
```

PR #550, opened 2019-10-10
Include version ids on response from field versions endpoint

lennym:
Because other comparison functions return a set of granted and latest changes even when the two are the same then the downstream consumer always expects the `latest` value to exist, and so when it's not returned it is still rendered, but as "No answer provided" because the value isn't sent.

Instead always return both granted and latest values, but also the version ids that those resolve to, so that downstream implementation can decide how best to deal with the case where those are the same.

### DIFF
--- a/pages/project-version/middleware/index.js
+++ b/pages/project-version/middleware/index.js
@@ -175,16 +175,13 @@ const getAllChanges = () => (req, res, next) => {
     getGrantedVersion(req)
   ])
     .then(([previousVersion, grantedVersion]) => {
-      return Promise.all([
-        getChanges(req.version, previousVersion),
-        getChanges(req.version, grantedVersion)
-      ]);
-    })
-    .then(([latest, granted]) => {
-      res.locals.static.changes = {
-        latest,
-        granted
+      return {
+        latest: getChanges(req.version, previousVersion),
+        granted: getChanges(req.version, grantedVersion)
       };
+    })
+    .then(changes => {
+      res.locals.static.changes = changes;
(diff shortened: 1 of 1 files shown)


pages/project-version/middleware/index.js
```diff
@@ -175,17 +175,14 @@ const getAllChanges = () => (req, res, next) => {
...
-        getChanges(req.version, grantedVersion)
-      ]);
-    })
-    .then(([latest, granted]) => {
-      res.locals.static.changes = {
-        latest,
-        granted
+      return {
+        latest: getChanges(req.version, previousVersion),
+        granted: getChanges(req.version, grantedVersion)
```
lennym: This bit is just removing `Promise.all` because the function isn't async so it was complicating things for no reason.

pages/project-version/middleware/index.js
```diff
@@ -196,18 +193,13 @@ const getChangedValues = (question, req) => {
...
-          granted: !isEqual(current, previous) && (previous || null)
-        };
-      }
       return {
-        previous: !isEqual(current, previous) && (previous || null),
-        granted: !isEqual(current, granted) && (granted || null)
+        grantedId: grantedVersion.id,
+        previousId: previousVersion.id,
+        previous,
+        granted
```
lennym: Also not sending the values when they hadn't changed was resulting in weirdness in the comparison window because it just assumed that the previous value was in fact `null`, and so rendered "No answer provided" for the before state, and not the actual value.

PR #552, opened 2019-10-10
Add route for editing comments

wheelsandcogs:


pages/project-version/middleware/index.js
```diff
@@ -10,34 +10,37 @@ const getVersion = () => (req, res, next) => {
     .catch(next);
 };
 
+const extractComments = task => {
```
lennym: Minor beef - this is in a middleware module and is not a middleware.

lennym: I'd create `./pages/project-version/lib/extract-comments.js` and put this function there.

wheelsandcogs: done

pages/project-version/lib/extract-comments.js
```diff
@@ -0,0 +1,25 @@
...
+      return {
+        id,
+        field: activity.event.meta.payload.meta.field,
+        comment,
+        deleted,
+        // we want to show the date of the following status change, not the comment submission.
+        createdAt: ([...statusChanges].reverse().find(s => s.createdAt > createdAt) || {}).createdAt,
+        author: `${firstName} ${lastName}`,
+        isNew
+      };
```
lennym: I know this is code that you've just moved from one place to another, but this destructuring to then immediately restructure is kinda wild.

I'm not going to beef for the purposes of merging this though.

PR #557, opened 2019-10-11
Feature/with admin status

joefitter:


pages/task/content/status.js
```diff
@@ -91,6 +91,24 @@ module.exports = {
...
+  // TODO: content
+  endorsed: {
+    state: 'Awaiting review',
+    action: {
+      application: 'Endorse application',
+      amendment: 'Endorse amendment'
+    },
+    hint: {
+      application: 'You confirm that the applicant holds the necessary training or experience to carry out the categories of procedures listed in this application.',
+      amendment: 'You confirm that the applicant holds the necessary training or experience to carry out the categories of procedures listed in this amendment.'
```
lennym: Can this be overwritten on a per-licence-type basis? It looks like this is _very_ PIL specific.

lennym: It'll be easier just to extend it in `./pages/task/read/content/project.js` I think. Assuming it works the way I think it does.

joefitter: I think it works like you think it does...

lennym: FWIW, since this is hint text, and therefore not _required_ rather than trying to write specific content I'd just set them to `''` for projects.

PR #558, opened 2019-10-14
Add amended date to project landing page

wheelsandcogs:


lennym:
This is going to be the last piece of the jigsaw so will need to sit here for a while.

wheelsandcogs:
You can merge this whenever as it just won't do anything until the amendDate prop is available.

PR #562, opened 2019-10-16
Updates to content

joefitter:
Updates to statuses and success messages for endorsement - also filtered end states from activity log if came from endorsedd

### DIFF
--- a/pages/task/read/views/models/project.jsx
+++ b/pages/task/read/views/models/project.jsx
@@ -81,20 +81,6 @@ export default function Project({ task, schema }) {
           {
             task.status === 'with-inspectorate' && (
               <Fragment>
-                <p><strong><Snippet>declarations.pel-holder.question</Snippet></strong></p>
-                <p>{declarations.authority}</p>
-                {
-                  declarationConfirmed(declarations.authority) &&
-                    <Fragment>
-                      <p>
-                        <Snippet>declarations.pel-holder.name</Snippet> {declarations['authority-pelholder-name']}
-                      </p>
-                      <p>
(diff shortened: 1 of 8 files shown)


pages/task/read/views/models/project.jsx
```diff
@@ -81,20 +81,6 @@ export default function Project({ task, schema }) {
...
-                  declarationConfirmed(declarations.authority) &&
-                    <Fragment>
-                      <p>
-                        <Snippet>declarations.pel-holder.name</Snippet> {declarations['authority-pelholder-name']}
-                      </p>
-                      <p>
-                        <Snippet>declarations.pel-holder.endorsement-date</Snippet> {declarations['authority-endorsement-date']}
-                      </p>
-                    </Fragment>
-                }
```
lennym: Can we leave this for now, for backwards compatibility reasons?

joefitter: Yes, yes we can

pages/task/read/views/models/project.jsx
```diff
@@ -84,15 +84,18 @@ export default function Project({ task, schema }) {
...
+                      <Fragment>
+                        <p>
+                          <Snippet>declarations.pel-holder.name</Snippet> {declarations['authority-pelholder-name']}
+                        </p>
+                        <p>
+                          <Snippet>declarations.pel-holder.endorsement-date</Snippet> {declarations['authority-endorsement-date']}
+                        </p>
+                      </Fragment>
+                    )
+                    : 'No answer provided'
```
lennym: I meant to put this in each of the two follow up questions.

`declarationConfirmed(declarations.authority)` will still return true if the application has been through a "new" endorsement journey, but the two subquestions will not have values.

PR #571, opened 2019-10-28
Change email improvements

wheelsandcogs:
* Better label for the password field
 * Early fail on entering current email as new email
 * Proper form error if new email is already in use

### DIFF
--- a/pages/user/update-email/routers/confirm.js
+++ b/pages/user/update-email/routers/confirm.js
@@ -1,5 +1,5 @@
 const { Router } = require('express');
-const { pick } = require('lodash');
+const { pick, set } = require('lodash');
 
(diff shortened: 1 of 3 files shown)


pages/user/update-email/routers/confirm.js
```diff
@@ -37,7 +37,13 @@ module.exports = () => {
         delete req.session.form[req.model.id];
       })
       .then(() => res.redirect(req.buildRoute('account.updateEmail.success')))
-      .catch(next);
+      .catch(err => {
+        if (err.message === 'Email address is already in use') {
```
lennym: 🤢 

There's _has to_ be a better way than that.

PR #577, opened 2019-10-31
Show more relevant content when asru amends a pil

wheelsandcogs:


pages/pil/update/index.js
```diff
@@ -90,10 +95,12 @@ module.exports = settings => {
...
   });
 
-  app.use('/success', success({
-    licence: 'pil',
-    status: 'resubmitted'
-  }));
+  app.use('/success', (req, res, next) => {
+    success({
+      licence: 'pil',
+      status: req.user.profile.asruLicensing ? 'autoresolved' : (req.user.profile.asruUser ? 'with-licensing' : 'resubmitted')
```
joefitter: I think this should be handled by looking at if there is an openTask and getting the details from there, otherwise the logic is spread across 2 places

wheelsandcogs: I was copying you :D
https://github.com/UKHomeOffice/asl-pages/blame/master/pages/pil/revoke/index.js#L23

PR #586, opened 2019-11-06
Respond with correct error code in permissions middleware

lennym:
If a user does not have permission to perform a task then the correct error code is a 403 - which is consistent with the response codes sent by the API layers.

### DIFF
--- a/lib/middleware/permissions.js
+++ b/lib/middleware/permissions.js
@@ -5,7 +5,7 @@ module.exports = (task, params) => (req, res, next) => {
       if (allowed) {
         return next();
       }
-      const err = new Error('Unauthorised');
+      const err = new Error('Forbidden');
       err.status = 403;
       throw err;
     })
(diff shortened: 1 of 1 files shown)


lib/middleware/permissions.js
```diff
@@ -6,7 +6,7 @@ module.exports = (task, params) => (req, res, next) => {
         return next();
       }
       const err = new Error('Unauthorised');
```
wheelsandcogs: Should this be "Forbidden" then?

lennym: If you're being _particularly_ anal then... maybe.

403 errors get handled with a bespoke error template anyway, so it doesn't matter. It depends how much of a stickler for the HTTP spec you are.

wheelsandcogs: well if we're going to be anal about using the correct error codes then we might as well send the accompanying message as well :)

wheelsandcogs:
```suggestion
      const err = new Error('Forbidden');
```

lennym: I'm more interested in "consistent" than "correct", but whatevs.

PR #587, opened 2019-11-06
Sort named people in PDF

wheelsandcogs:


pages/establishment/pdf/views/establishment.jsx
```diff
@@ -1,10 +1,13 @@
 import React, { Fragment } from 'react';
 import ReactMarkdown from 'react-markdown';
 import map from 'lodash/map';
+import sortBy from 'lodash/sortBy';
 import StandardConditions from '../../../common/views/pdf/standard-conditions';
 
 const getName = person => `${person.firstName} ${person.lastName}`;
 
+const getPeopleList = people => sortBy(people, ['lastName']).map(getName).join(', ');
```
lennym: On the off chance there are people with the same surname.

```suggestion
const getPeopleList = people => sortBy(people, ['lastName', 'firstName']).map(getName).join(', ');
```

PR #589, opened 2019-11-08
Species selector for pils

joefitter:
* bumped components and react components with species selector field type
* add new step after training modules for species
* updated species joining logic for top level pil
* show species on task reviewn


pages/pil/training/routers/species.js
```diff
@@ -0,0 +1,53 @@
...
+    return req.api(`/establishment/${req.establishmentId}/profiles/${req.profileId}/certificate`, opts)
+      .then(() => {
+        delete req.session.form[req.model.id];
+        return next();
+      })
+      .catch(next);
+  });
+
+  app.post('/', (req, res, next) => {
+    return res.redirect(req.buildRoute('pil.update'));
```
wheelsandcogs: Any reason you separate this rather than just calling it after `delete req.session.form[req.model.id];` ?

joefitter: yes, convention - having lots of middlewares is a good thing, it makes it clearer for me anyway.

joefitter:
Wipping as species selector should only be shown if modules require species

PR #600, opened 2019-11-21
Feature - refactor routes

joefitter:
* Added route config to all top level components
* Updated link page paths
* Use suffix in buildRoute

Waiting for https://github.com/UKHomeOffice/asl-service/pull/166 and https://github.com/UKHomeOffice/asl-components/pull/92


pages/common/components/licence-status-banner.jsx
```diff
@@ -0,0 +1,89 @@
...
+      </div>
+    );
+  }
+}
+
+LicenceStatusBanner.defaultProps = {
+  dateFormat: 'DD MMMM YYYY'
+};
+
+export default LicenceStatusBanner;
```
joefitter: moved here from asl/components - to be deleted from there once merged

PR #607, opened 2019-11-22
Transfer a PIL

wheelsandcogs:


pages/common/assets/sass/style.scss
```diff
@@ -316,3 +316,19 @@ p.spoc {
...
+    border-top: 1px solid govuk-colour('grey-2');
+    padding: $govuk-gutter 0;
+  }
+}
+
+details {
+  margin-bottom: $govuk-gutter;
+
+  summary {
+    color: #005ea5;
```
joefitter: is this `$govuk-link-colour`? if so could we use the variable, or if it's a new thing could we define it at the top of the file

pages/pil/dashboard/index.js
```diff
@@ -39,6 +47,16 @@ module.exports = settings => {
   app.use((req, res, next) => {
     const values = get(req.session, `form[${req.model.id}].values`);
     req.model = { ...req.model, ...values };
+
+    const establishmentTransfer = req.user.profile.establishments
+      .filter(e => e.id !== req.establishment.id)
+      .find(e => e.id === get(req.model, 'establishmentId'));
+
+    req.model.establishment = {
+      from: pick(req.establishment, ['id', 'name']),
```
joefitter: This feels like it should be taken from the PIL model, req.pil.establishment - I know with current business logic it does also happen to be the establishment you are scoped to in the URL but if that changes in future, or if we allow someone with perms at 2 establishments to view this pil then this logic becomes broken. I'd personally eager load the establishment at API level

pages/pil/dashboard/index.js
```diff
@@ -84,6 +102,9 @@ module.exports = settings => {
       res.locals.static.skipTraining = get(req.session, [req.profileId, 'skipTraining'], null);
       res.locals.static.isAsru = req.user.profile.asruUser;
       res.locals.static.isLicensing = req.user.profile.asruLicensing;
+      res.locals.static.canTransferPil = req.pil.status === 'active' && req.user.profile.id === req.profile.id; // can only transfer own (active) pil
+
+      // res.locals.static.canTransferPil = false; // PIL TRANSFER FEATURE FLAG (disables edit establishment link)
```
joefitter: assume this will be removed before this PR is un-WIPed?

pages/task/read/content/pil.js
```diff
@@ -30,12 +31,15 @@ module.exports = {
     applicant: {
       application: 'Applicant',
       amendment: 'Licence holder',
-      revocation: 'Licence holder'
+      revocation: 'Licence holder',
+      transfer: 'Licence holder'
     },
+    // establishment: 'Establishment transfer',
```
joefitter: is this to be removed?

joefitter:
```suggestion
```

pages/pil/dashboard/index.js
wheelsandcogs: @lennym requested that we had an easy way to disable the UI for this feature so that we can still test it on prod without having the link actually visible - I've commented it out as we might want to leave it on for S&T demo if it makes the cut?

pages/pil/dashboard/index.js
```diff
@@ -99,6 +120,7 @@ module.exports = settings => {
   app.get('/success', (req, res, next) => {
     success({
       licence: 'pil',
+      type: get(req.model, 'openTasks[0].data.action') === 'transfer' ? 'transfer' : null,
```
lennym: What's this for?

lennym: Or more specifically, why is it important to send specifically `null` in the negative case and not just the task action all the time?

pages/pil/dashboard/index.js
joefitter: OIC - carry on then

pages/pil/dashboard/index.js
wheelsandcogs: The success router handles fallback for application vs amendment if you don't pass it a value, didn't want to replicate that logic here.

lennym: It feels like we're spreading that logic around, and possibly the logic to determine the type of the thing should be handled all in one place.

pages/common/routers/success.js
```diff
@@ -17,8 +17,14 @@ module.exports = ({
...
-    if (!type && req.model && req.model.status) {
-      type = req.model.status === 'active' ? 'amendment' : 'application';
+    if (!type) {
+      if (get(req.model, 'openTasks[0].data.action') === 'transfer') {
+        type = 'transfer';
+      } else if (get(req.model, 'status') === 'active') {
+        type = 'amendment';
+      } else {
+        type = 'application';
+      }
```
lennym: 👍

pages/pil/dashboard/index.js
```diff
@@ -84,6 +110,9 @@ module.exports = settings => {
       res.locals.static.skipTraining = get(req.session, [req.profileId, 'skipTraining'], null);
       res.locals.static.isAsru = req.user.profile.asruUser;
       res.locals.static.isLicensing = req.user.profile.asruLicensing;
+      res.locals.static.canTransferPil = req.pil.status === 'active' && req.user.profile.id === req.profile.id; // can only transfer own (active) pil
+
+      // res.locals.static.canTransferPil = false; // PIL TRANSFER FEATURE FLAG (disables edit establishment link)
```
lennym:
```suggestion
```

PR #619, opened 2019-11-26
Bugfix - reveal validation

joefitter:
* Update form router to validate relevant option reveals
* Pass errors to connected child component

### DIFF
--- a/pages/common/routers/form.js
+++ b/pages/common/routers/form.js
@@ -13,7 +13,8 @@ const {
   pick,
   omit,
   chain,
-  reduce
+  reduce,
+  castArray
 } = require('lodash');
 const validator = require('../../../lib/validation');
 const { hasChanged, cleanModel } = require('../../../lib/utils');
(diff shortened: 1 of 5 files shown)


lennym:
Any chance of some unit tests for the stuff in the form router?

joefitter:
YES, #drunkentraincoding doesn't come with any guarantees

pages/common/routers/form.js
```diff
@@ -27,6 +28,22 @@ if (process.env.CSRF === 'false') {
...
+    if (!field.options) {
+      return obj;
+    }
+    const selectedOptions = field.options.filter(opt => castArray(values[key]).includes(opt.value));
+    return selectedOptions.reduce((o, opt) => {
+      return {
+        ...o,
+        ...(opt.reveal || {})
+      };
+    }, {});
```
lennym: There's no mention of `obj` here, which makes me suspect that this won't work if there's more than one field in the schema which has reveal options.

I _think_ it needs this:

```suggestion
    }, obj);
```

lennym: Narrator voice: It did need that.

PR #620, opened 2019-11-27
Add a cancel button to the edit PIL establishment form

wheelsandcogs:


pages/pil/establishment/index.js
```diff
@@ -26,6 +26,10 @@ module.exports = settings => {
     configure: (req, res, next) => {
       req.form.schema = getSchema(req.user.profile.establishments, req.pil);
       next();
+    },
+    cancelEdit: (req, res, next) => {
```
lennym: Is there some "magic" in the form router that hits this?

wheelsandcogs: yup, we use it in a few places https://github.com/UKHomeOffice/asl-pages/blob/master/pages/common/routers/README.md#usage

PR #621, opened 2019-11-28
Feature - Invitation actions

joefitter:
* Added resend cancel/delete options to invitations
* Added success notifications

waiting for https://github.com/UKHomeOffice/asl-resolver/pull/118
and
https://github.com/UKHomeOffice/asl-public-api/pull/196

### DIFF
--- a/pages/profile/invitations/index.js
+++ b/pages/profile/invitations/index.js
@@ -1,6 +1,6 @@
 const { merge } = require('lodash');
 const { page } = require('@asl/service/ui');
-const { BadRequestError } = require('@asl/service/errors');
+const { NotFoundError } = require('@asl/service/errors');
 const datatable = require('../../common/routers/datatable');
 const content = require('./content');
 const schema = require('./schema');
--- a/pages/profile/invitations/index.js
+++ b/pages/profile/invitations/index.js
@@ -1,5 +1,8 @@
+const { merge } = require('lodash');
 const { page } = require('@asl/service/ui');
+const { BadRequestError } = require('@asl/service/errors');
 const datatable = require('../../common/routers/datatable');
+const content = require('./content');
 const schema = require('./schema');
 
 module.exports = settings => {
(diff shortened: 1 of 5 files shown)


pages/profile/invitations/index.js
```diff
@@ -25,5 +28,27 @@ module.exports = settings => {
...
+    if (!allowedActions.includes(req.params.action)) {
+      return next(new BadRequestError('Unrecognised action'));
+    }
+
+    const { invitationId, action } = req.params;
+    const isDeletion = action === 'delete';
+    const params = {
+      method: isDeletion ? 'DELETE' : 'PUT'
+    };
+    const suffix = isDeletion ? '' : `/${action}`;
```
lennym:
```suggestion
    const suffix = isDeletion ? '' : `/delete`;
```

We've just asserted that `action === 'delete'` so might as well be explicit about what's going on.

pages/profile/invitations/index.js
```diff
@@ -1,5 +1,8 @@
+const { merge } = require('lodash');
 const { page } = require('@asl/service/ui');
+const { BadRequestError } = require('@asl/service/errors');
```
lennym:
```suggestion
const { NotFoundError } = require('@asl/service/errors');
```

pages/profile/invitations/index.js
```diff
@@ -25,5 +28,27 @@ module.exports = settings => {
     }
   })({ schema }));
 
+  app.post('/:invitationId/:action', (req, res, next) => {
+    const allowedActions = ['delete', 'resend', 'cancel'];
+
+    if (!allowedActions.includes(req.params.action)) {
+      return next(new BadRequestError('Unrecognised action'));
```
lennym:
```suggestion
      return next(new NotFoundError());
```

pages/profile/invitations/index.js
joefitter: that doesn't make any sense :(

joefitter: if its a DELETE we _dont_ want a suffix

joefitter: we just DELETE /:id

lennym: My bad.

PR #631, opened 2019-12-04
Feature - Remove declarations if ASRU changes SOP

joefitter:
* Removed declaration if ASRU for create update and delete
* Fixed copy on delete role confirm page - submit to ASRU -> submit

### DIFF
--- a/pages/place/index.js
+++ b/pages/place/index.js
@@ -6,7 +6,10 @@ module.exports = settings => {
   const app = Router({ mergeParams: true });
 
   app.param('placeId', (req, res, next, placeId) => {
-    return req.api(`/establishment/${req.establishmentId}/place/${placeId}`)
+    const params = req.path.match(/delete\/success$/)
+      ? { query: { withDeleted: true } }
+      : {};
+    return req.api(`/establishment/${req.establishmentId}/place/${placeId}`, params)
       .then(({ json: { data, meta } }) => {
         req.placeId = placeId;
         res.locals.static.establishment = meta.establishment;
--- a/pages/place/routers/confirm.js
+++ b/pages/place/routers/confirm.js
@@ -27,46 +27,49 @@ module.exports = settings => {
 
   app.post('/', updateDataFromTask(sendData));
 
-  app.use(form(Object.assign({
-    model: 'place',
-    schema: declarationsSchema,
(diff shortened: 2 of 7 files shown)


joefitter:
needs https://github.com/UKHomeOffice/asl-workflow/pull/190 to be deployed at the same time

pages/place/routers/confirm.js
```diff
@@ -27,46 +27,49 @@ module.exports = settings => {
...
-    },
-    editAnswers: (req, res, next) => {
-      delete req.session.form[req.model.id].validationErrors;
-      return res.redirect(req.buildRoute(settings.page));
-    },
-    cancelEdit: (req, res, next) => {
-      return res.redirect(req.buildRoute('place.list'));
-    }
-  }, settings)));
+    }, settings))(req, res, next);
```
lennym: I really hate the pattern of middlewares that call other middlewares.

Could we set `req.requiresDeclaration` in a previous step, and get rid of that?

lennym: Alternatively make `requiresDeclaration` a function that takes the request as an argument and returns a `Boolean`

joefitter: don't be a hater

joefitter: also yes, although the other 2 corresponding PRs do it this way, so I guess you want them updating too?

pages/role/remove/content/confirm.js
```diff
@@ -12,7 +12,7 @@ module.exports = merge({}, baseContent, {
     declaration1: `Applicant authority, conflict of interests and PELh authority declaration text to go here`
   },
   buttons: {
-    submit: 'Submit to ASRU'
+    submit: 'Submit'
```
lennym: 🎉

pages/place/index.js
```diff
@@ -6,7 +6,10 @@ module.exports = settings => {
   const app = Router({ mergeParams: true });
 
   app.param('placeId', (req, res, next, placeId) => {
-    return req.api(`/establishment/${req.establishmentId}/place/${placeId}`)
+    const params = req.path.match(/delete\/success$/)
+      ? { query: { withDeleted: true } }
+      : {};
```
lennym: I'd be tempted to just hard-code this.

PR #633, opened 2019-12-04
Apply for PEL

wheelsandcogs:


pages/establishment/dashboard/views/index.jsx
```diff
@@ -49,11 +50,22 @@ const Index = ({
           <PanelList
             panels={links.filter(link => allowedActions.includes(link.permissions)).map((link, index) => <DashboardLink key={index} { ...link } />)}
           />
+          {
+            establishment.status !== 'active' && allowedActions.includes('establishment.update') && !openApplication &&
```
lennym: This is a lot of conditions to have inline. Could we pull it out into a `const canApply = establishment.status !== 'active' && allowedActions.includes('establishment.update') && !openApplication;` ?

pages/common/middleware/index.js
```diff
@@ -70,9 +70,29 @@ const clearSessionIfNotFromTask = () => (req, res, next) => {
...
+
+    if (pelh) {
+      req.establishment.pelh = pelh.profile;
+    }
+    if (nprc && (!pelh || nprc.profile.id !== pelh.profile.id)) {
+      req.establishment.nprc = nprc.profile;
+    }
+
+    ['holc', 'nacwo', 'nio', 'nvs', 'ntco'].map(roleName => {
+      req.establishment[roleName] = get(req.establishment.roles.find(r => r.type === roleName), 'profile');
```
lennym: This doesn't seem right - it's going to return only the first person in a role. What if there are multiple people holding the same role?

lennym: Note: for NACWOs in particular that's highly likely to be the case.

wheelsandcogs: Yeah I forgot there can be more than one of each, will fix.

pages/establishment/dashboard/views/index.jsx
```diff
@@ -38,6 +38,8 @@ const Index = ({
 }) => {
   const inspectors = establishment.asru.filter(p => p.asruInspector);
   const spocs = establishment.asru.filter(p => p.asruLicensing);
+  const openApplication = establishment.openTasks.find(task => task.data.model === 'establishment' && task.data.action === 'grant');
```
lennym:
```suggestion
  const openApplication = establishment.openTasks.find(task => task.data.model === 'establishment' && allowedActions.includes('establishment.update') && task.data.action === 'grant');
```

pages/establishment/dashboard/views/index.jsx
```diff
@@ -49,20 +51,44 @@ const Index = ({
           <PanelList
             panels={links.filter(link => allowedActions.includes(link.permissions)).map((link, index) => <DashboardLink key={index} { ...link } />)}
           />
+          {
+            canApply &&
+              <Link page="establishment.apply" label={<Snippet>buttons.establishment.apply</Snippet>} className="govuk-button" />
+          }
+          {
+            allowedActions.includes('establishment.update') && openApplication &&
```
lennym:
```suggestion
            openApplication &&
```

lennym: Since the other rendering condition rolls in a permissions check it makes sense to do the same for this one.

PR #635, opened 2019-12-06
Refactor - remove application confirm

joefitter:
* Application confirm component is now available as a field type, added optional function in form router to append this field if truthy
* Fixed bug in establishment details update task - we were ignoring the 3 flag fields as mapping happened before hyrdration
* removed custom <form> tags on confirm pages


pages/establishment/dashboard/views/index.jsx
```diff
@@ -83,7 +83,7 @@ const Index = ({
                     {
                       establishment.holc.map(holc => (
                         <p key={holc.id} className="holc">
-                          <Link page="globalProfile" profileId={holc.id} label={`${holc.firstName} ${holc.lastName}`} />
+                          <Link page="profile.read" profileId={holc.id} label={`${holc.firstName} ${holc.lastName}`} />
```
lennym: 👍

PR #639, opened 2019-12-09
Don't overwrite pseudo form fields with empty session data

lennym:
The `getValue` functions are being invoked for "pseudo" form fields even when there is no session data present because the session is initialised with a value of `{}` which is always truthy.

Instead check that the session is non-empty before extending the model values with any session values.


wheelsandcogs:
2 lines of fix and 55 of tests 👍

PR #644, opened 2019-12-12
Use pdf perms for pil, pass correct params, use establisment from pil…

joefitter:
…, not req

needs
https://github.com/UKHomeOffice/asl-permissions/pull/48 
and
https://github.com/UKHomeOffice/asl-service/pull/174

to unwip

### DIFF
--- a/pages/pil/routes.js
+++ b/pages/pil/routes.js
@@ -35,7 +35,7 @@ module.exports = {
   },
   pdf: {
     path: '/:pilId/pdf',
-    permissions: 'pil.read',
+    permissions: 'pil.pdf',
     router: pdf
   }
 };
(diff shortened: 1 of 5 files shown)


pages/pil/routes.js
```diff
@@ -1,10 +1,20 @@
...
 const revoke = require('./revoke');
 const update = require('./update');
 const remove = require('./delete');
 const pdf = require('./pdf');
 const read = require('./read');
 const create = require('./create');
 
+const pilPermissions = task => (req, res, next) => {
+  const params = {
+    id: req.pilId,
```
joefitter:
```suggestion
    id: req.pil.id,
```

pages/pil/routes.js
```diff
@@ -1,10 +1,20 @@
...
 const remove = require('./delete');
 const pdf = require('./pdf');
 const read = require('./read');
 const create = require('./create');
 
+const pilPermissions = task => (req, res, next) => {
+  const params = {
+    id: req.pil.id,
+    profileId: req.pil.profileId,
+    establishment: req.pil.establishmentId
```
joefitter:
```suggestion
    // this is to ensure only users who have permissions at the holding establishment can print the PDF
    establishment: req.pil.establishmentId
```

PR #650, opened 2019-12-18
Remove the "This item is currently being edited" message when edit and resubmitting a PIL

wheelsandcogs:


joefitter:
I don't think just removing the message is sufficient, is there a reason it is being removed without blocking the PIL from being edited (without going through the task)?

wheelsandcogs:
Closing until we get more clarity on the ticket

PR #653, opened 2019-12-27
Feature - billing

joefitter:
* Added routes for establishment scoped billing pages
* Added datatable for relevant pils
* Added summary page


pages/common/assets/sass/style.scss
```diff
@@ -28,6 +28,20 @@
...
+  .pull-right {
+    width: 15%;
+  }
+}
+
+p.subtitle {
+  font-size: 1.1em;
+}
+
+.margin-bottom {
```
wheelsandcogs: this feels a bit icky

pages/common/content/index.js
```diff
@@ -9,7 +9,8 @@ module.exports = {
       dashboard: '{{establishment.name}}',
       read: 'Details',
       update: 'Amend establishment details',
-      apply: 'Apply for licence'
+      apply: 'Apply for licence',
+      fees: 'Projected licence fees'
```
wheelsandcogs:
```suggestion
      fees: 'Estimated licence fees'
```
In the S&T they asked not to use "Projected" to avoid confusion with Projects.

wheelsandcogs: "Forecasted" might work also to avoid confusion between Estimated / Establishment 😄

pages/establishment/licence-fees/content/index.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
+  fees: {
+    title: 'Projected licence fees',
```
wheelsandcogs:
```suggestion
    title: 'Estimated licence fees',
```

pages/establishment/licence-fees/personal/formatters/index.jsx
```diff
@@ -0,0 +1,27 @@
...
+export default {
+  licenceHolder: {
+    format: (val, pil) => `${pil.profile.firstName} ${pil.profile.lastName}`
+  },
+  licenceNumber: {
+    format: (val, pil) => {
+      if (!pil.profile.establishments.length) {
+        return val;
+      }
+      return <Link page="pil.read" pilId={pil.id} profileId={pil.profile.id} label={val} establishmentId={pil.establishmentId} />;
```
wheelsandcogs: Are we linking to establishments that the current user might not be related to, or are they stripped elsewhere?

lennym: It is stripped elsewhere, but it's redundant here, so I've removed it anyway.

pages/common/assets/sass/style.scss
lennym: I don't disagree, but also I'm not inclined to go and find every instance of a details element to work out how doing it differently would affect things.

pages/common/content/index.js
lennym: Will change these in the future once someone has taken a position on what it _should_ be. It's going to be a long way off from being visible to external users anyway.

PR #658, opened 2020-01-13
Add the shared key to the establishment sidebar

wheelsandcogs:


pages/establishment/dashboard/views/index.jsx
```diff
@@ -147,6 +147,14 @@ const Index = ({
...
               }
             </dd>
+
+            {
+              establishment.sharedKey &&
+                <Fragment>
+                  <dt><Snippet>sharedKey</Snippet></dt>
+                  <dd>{establishment.sharedKey}</dd>
+                </Fragment>
+            }
```
lennym: Should this be visible to _everyone_ who has visibility of the establishment?

wheelsandcogs: Nope, just ASRU, I'll fix.

lennym: Can it be fixed in a way that reads the permissions from permissions rather than baking this particular permission into the UI?

PR #661, opened 2020-01-16
Add UI and endpoints for PIL fee waivers

lennym:


pages/establishment/licence-fees/personal/views/row.jsx
```diff
@@ -4,16 +4,23 @@ import fetch from 'r2';
...
+
+function UpdatingForm({ id, waived, onCancelClick, formFields }) {
   return (
     <Fragment>
       {
         formFields
       }
       <input type="hidden" name="pilId" value={id} />
-      <input type="hidden" name="billable" value={billable} />
+      <input type="hidden" name="waived" value={!waived} />
```
wheelsandcogs: This feels a bit weird but maybe I'm just misunderstanding it? We submit a value of waived but it's flipped from the value that was loaded?

lennym: Yes, because the form is to turn it on and off, so it needs to flip somewhere.

wheelsandcogs: Ah ok, didn't realise this was the toggle.

PR #662, opened 2020-01-16
Add CSV export functions to datatable router

lennym:
Allows downloading a CSV version of the data in a table by appending a `?csv=1` to the page's query string.

### DIFF
--- a/pages/common/routers/datatable.js
+++ b/pages/common/routers/datatable.js
@@ -84,6 +84,9 @@ module.exports = ({
       }
     }
     let { rows = defaultRowCount, page } = req.query;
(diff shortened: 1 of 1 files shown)


pages/common/routers/datatable.js
```diff
@@ -84,6 +84,9 @@ module.exports = ({
       }
     }
     let { rows = defaultRowCount, page } = req.query;
+    if (req.query.csv) {
+      rows = 1e6;
```
wheelsandcogs: What's this doing?

wheelsandcogs: 1 mil?

PR #669, opened 2020-01-22
Hide the countdown unless the licence is active

wheelsandcogs:
Needs https://github.com/UKHomeOffice/asl-components/pull/103 bumping first

### DIFF
--- a/pages/project/formatters/index.jsx
+++ b/pages/project/formatters/index.jsx
@@ -30,7 +30,12 @@ const formatters = {
       <Link page="profile.read" profileId={id} label={`${firstName} ${lastName}`} />
   },
   expiryDate: {
-    format: date => date ? <ExpiryDate date={date} showNotice={11} /> : '-'
+    format: (date, model) => {
+      if (!date) {
+        return '-';
+      }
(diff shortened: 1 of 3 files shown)


wheelsandcogs:
Technically this isn't required at the minute, as we only show the expiry date column in the "Active" projects tab externally, so there's never a case where the model isn't active.

However, we use a similar formatter in `asl-internal-ui` (where this fix is applied) so they should probably behave the same to avoid any future confusion or bugs if we add the expiry date column for non-active-projects in external ui.

Related PR: https://github.com/UKHomeOffice/asl-internal-ui/pull/151

pages/project/formatters/index.jsx
```diff
@@ -30,7 +30,12 @@ const formatters = {
       <Link page="profile.read" profileId={id} label={`${firstName} ${lastName}`} />
   },
   expiryDate: {
-    format: date => date ? <ExpiryDate date={date} showNotice={11} /> : '-'
+    format: (date, model) => {
+      if (!date) {
+        return '-';
+      }
+      return <ExpiryDate date={date} showNotice={model.status === 'active' ? 11 : false} />;
```
lennym: ELEVEN!!

PR #671, opened 2020-01-23
Allow ASRU to initiate PIL transfers

wheelsandcogs:
Needs https://github.com/UKHomeOffice/asl-public-api/pull/213

### DIFF
--- a/pages/pil/dashboard/index.js
+++ b/pages/pil/dashboard/index.js
@@ -51,7 +51,7 @@ module.exports = settings => {
     const values = get(req.session, `form[${req.model.id}].values`);
     req.model = { ...req.model, ...values };
 
-    const establishmentTransfer = req.user.profile.establishments
+    const establishmentTransfer = req.profile.establishments
       .filter(e => e.id !== req.establishment.id)
       .find(e => e.id === get(req.model, 'establishmentId'));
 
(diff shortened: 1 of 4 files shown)


pages/pil/dashboard/index.js
```diff
@@ -91,18 +91,22 @@ module.exports = settings => {
...
+        .then(() => next())
+        .catch(next);
     }
   }));
 
   app.post('/', redirectToTaskIfOpen());
 
   app.post('/', (req, res, next) => {
     sendData(req)
+      .then(() => delete req.session.form[req.model.id])
```
lennym: What's this for?

wheelsandcogs: We were not clearing the form after submission so if you went back into it it was persisting the values

pages/pil/dashboard/index.js
```diff
@@ -91,18 +91,22 @@ module.exports = settings => {
...
-      // can only transfer own pil if it's active and has no in-progress amendment
-      const hasOpenAmendment = req.pil.status === 'active' && get(req.model, 'openTasks[0].data.action') === 'grant';
-      res.locals.static.canTransferPil = req.pil.status === 'active' && req.user.profile.id === req.profile.id && !hasOpenAmendment;
+      return req.user.can('pil.transfer', { pilId: req.pilId })
+        .then(canTransfer => {
+          const hasOpenAmendment = req.pil.status === 'active' && get(req.model, 'openTasks[0].data.action') === 'grant';
 
-      next();
+          res.locals.static.canTransferPil = req.pil.status === 'active' && !hasOpenAmendment && canTransfer;
+        })
```
lennym: I'd be tempted to switch the order of these checks so that the call to `req.user.can` is only executed once the local checks have passed. This means we can avoid the extra API call in the case where we already know it's not transferable.

wheelsandcogs: I had it that way initially for exactly that reason but decided the tradeoff in better readability was worth the extra call. I can switch it back.

lennym: It might just be the diff view, but it's not _super_ readable as it is, so I'd switch it back.

Might be worth pulling out into it's own function, especially since it looks a lot like there's a similar (but not identical) bit of logic in the `pil/establishment` router too 👇

pages/pil/establishment/index.js
```diff
@@ -15,16 +15,18 @@ module.exports = settings => {
...
-      throw new Error('You can only transfer your own PIL');
-    }
-    next();
+    return req.user.can('pil.transfer', { pilId: req.pilId })
+      .then(canTransfer => {
+        if (!canTransfer) {
+          return next(new Error('Only the PIL holder and ASRU can transfer this PIL'));
+        }
+      })
+      .then(() => next());
```
lennym: Here.

pages/pil/dashboard/index.js
lennym: 👍

lennym:
It would be good to see this get some integration test coverage once it's all shipped.

PR #680, opened 2020-01-29
Remove grey background from expanding panels inside panel lists

lennym:
As seen on the external home page and global profile pages. These random borders and backgrounds look super nasty and should be removed.

Before:

Screenshot 2020-01-29 at 12 27 39

After:

Screenshot 2020-01-29 at 12 28 03


joefitter:
🎉

PR #681, opened 2020-01-29
Feature - restrictions improvements

joefitter:
* Replaced conditional reveal and text with new restrictions field type for place edits
* Replaced Field with EditableField in place task review.
* Added warning if user navigates away from task after changing restrictions and not progressing the task

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "@asl/components": {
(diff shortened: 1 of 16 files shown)


joefitter:
WIP until https://github.com/UKHomeOffice/asl-public-api/pull/214 and https://github.com/UKHomeOffice/asl-components/pull/107 merged

PR #706, opened 2020-02-17
Use a more descriptive name for project status banner

wheelsandcogs:


pages/project-version/downloads/views/index.jsx
```diff
@@ -1,7 +1,7 @@
 import React, { Fragment } from 'react';
 import { useSelector } from 'react-redux';
 import { Snippet } from '@asl/components';
-import LicenceStatusBanner from '../../components/status-banner';
+import ProjectStatusBanner from '../../components/status-banner';
```
lennym: Can we change the file name too?

PR #708, opened 2020-02-18
Refactor pdf to use shared licence status banner

wheelsandcogs:


pages/common/views/pdf/header.jsx
```diff
@@ -1,43 +1,55 @@
...
-          <Fragment> | Revoked: <strong>{format(model.revocationDate)}</strong></Fragment>
-        }
-      </p>
-    </header>
+          { model.status === 'revoked' &&
+            <Fragment> | Revoked: <strong>{format(model.revocationDate)}</strong></Fragment>
+          }
+        </p>
+      </header>
+    </Provider>
```
lennym: It's a bit out of scope here, but it seems like a lot of the things in this file are switching on and off depending on whether it's a PPL or not. There's probably a case for just splitting the whole thing into two components with minimal logic.

PR #710, opened 2020-02-19
Bugfix - content mismatches

joefitter:
* control panel doesn't need to be a p
* fixed various content mismatch errors
* dont put a form in a p

### DIFF
--- a/pages/establishment/licence-fees/views/index.jsx
+++ b/pages/establishment/licence-fees/views/index.jsx
@@ -21,7 +21,10 @@ function Fee({ type }) {
   return (
     <div>
       <h3><Snippet>{`fees.summary.${type}`}</Snippet></h3>
-      <h1>£{numberWithCommas(num)}</h1>
+      <h1>
+        {'£'}
+        <span>{ numberWithCommas(num) }</span>
+      </h1>
     </div>
   );
 }
--- a/pages/project/list/views/index.jsx
+++ b/pages/project/list/views/index.jsx
@@ -35,12 +35,12 @@ const Projects = ({
     />
     {
       allowedActions.includes('project.apply') && (
-        <p className="control-panel projects-actions">
+        <div className="control-panel projects-actions">
           <form method="POST" action="projects/create">
             <Button><Snippet>buttons.create</Snippet></Button>
           </form>
           <Link page="project.import" className="govuk-button button-secondary" label={<Snippet>buttons.upload</Snippet>} />
-        </p>
+        </div>
       )
     }
     <Tabs active={tabs.indexOf(status)}>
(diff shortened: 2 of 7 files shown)


pages/establishment/licence-fees/views/index.jsx
```diff
@@ -21,7 +21,10 @@ function Fee({ type }) {
   return (
     <div>
       <h3><Snippet>{`fees.summary.${type}`}</Snippet></h3>
-      <h1>£{numberWithCommas(num)}</h1>
+      <h1>
+        {'£'}
```
lennym: :-|

pages/project/list/views/index.jsx
```diff
@@ -35,12 +35,12 @@ const Projects = ({
     />
     {
       allowedActions.includes('project.apply') && (
-        <p className="control-panel projects-actions">
+        <div className="control-panel projects-actions">
```
lennym: This will likely have knock-on effects on styling because a `p` comes with inherent styles but a `div` does not.

pages/project/read/views/index.jsx
```diff
@@ -309,7 +309,7 @@ export default function ProjectLandingPage() {
 
         <dt><Snippet>fields.licenceHolder.label</Snippet></dt>
         <dd>
-          {model.licenceHolder.firstName} {model.licenceHolder.lastName}<br />
+          {model.licenceHolder.name}<br />
```
lennym: I would be wary about introducing dependency on the virtual property `name` existing. It's not there in certain contexts and should probably be removed generally.

If there's a reason why the old way didn't work then I'd go for this:

```suggestion
          {`${model.licenceHolder.firstName} ${model.licenceHolder.lastName}`}<br />
```

pages/task/read/views/components/activity-log.jsx
```diff
@@ -27,10 +27,10 @@ const getAuthor = ({ changedBy, eventName, event: { status } }, task) => {
   }
 
   const name = `${changedBy.firstName} ${changedBy.lastName}`;
-  const action = <Snippet fallback={`status.${status}.log`}>{`status.${status}.log.${task.type}`}</Snippet>;
   return (
     <p>
-      <strong>{action}: </strong>
+      <strong><Snippet fallback={`status.${status}.log`}>{`status.${status}.log.${task.type}`}</Snippet></strong>
+      <strong>: </strong>
```
lennym: React is a dick.

pages/project/list/views/index.jsx
joefitter: it does not - I checked :)

pages/task/read/views/components/activity-log.jsx
joefitter: yes.

PR #715, opened 2020-02-20
Digitise paper project licences

wheelsandcogs:


pages/project/read/views/index.jsx
```diff
@@ -354,7 +387,7 @@ export default function ProjectLandingPage() {
                 </Fragment>
               }
               {
-                !isRevoked &&
+                model.expiryDate && !isRevoked &&
```
lennym: What's the reasoning for adding this? How is it possible to have a granted licence with no expiry date?

wheelsandcogs: I added this before we had a duration field in the initial form to create the stub, so there was no expiry date.

Now that we require a duration, it's a bit redundant - should I remove it?

lennym: Yes please. If a project manages to get into this state then it’s probably better if it errors and alerts.

PR #716, opened 2020-02-27
Check roles and projects have correct establishment when removing permissions

lennym:
If a user has roles and projects at any establishments then they can't be removed from other establishments.

Make sure that roles and projects have establishment ids that match the current estasblishment when removing permissions.

### DIFF
--- a/pages/profile/permission/index.js
+++ b/pages/profile/permission/index.js
@@ -20,10 +20,10 @@ module.exports = settings => {
   }));
 
   app.use('/', (req, res, next) => {
-    const hasRoles = !!(req.profile.roles && req.profile.roles.length);
-    const hasPil = !!(req.profile.pil && req.profile.pil.status === 'active' && req.profile.pil.establishmentId === req.establishmentId);
-    const hasProjects = !!(req.profile.projects && req.profile.projects.length && some(req.profile.projects, project => project.status === 'active'));
-    res.locals.static.isNamed = hasRoles || hasPil || hasProjects;
+    const hasRoles = some(req.profile.roles, role => role.establishmentId === req.establishmentId);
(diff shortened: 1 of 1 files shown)


pages/profile/permission/index.js
```diff
@@ -20,10 +20,10 @@ module.exports = settings => {
...
 
   app.use('/', (req, res, next) => {
-    const hasRoles = !!(req.profile.roles && req.profile.roles.length);
-    const hasPil = !!(req.profile.pil && req.profile.pil.status === 'active' && req.profile.pil.establishmentId === req.establishmentId);
-    const hasProjects = !!(req.profile.projects && req.profile.projects.length && some(req.profile.projects, project => project.status === 'active'));
-    res.locals.static.isNamed = hasRoles || hasPil || hasProjects;
+    const hasRoles = some(req.profile.roles, role => role.establishmentId === req.establishmentId);
+    const hasPil = req.profile.pil && req.profile.pil.status === 'active' && req.profile.pil.establishmentId === req.establishmentId;
+    const hasProjects = some(req.profile.projects, project => project.status === 'active' && project.establishmentId === req.establishmentId);
+    res.locals.static.isNamed = !!hasRoles || !!hasPil || !!hasProjects;
```
wheelsandcogs: Aren't they all booleans already, so no need to cast here?

lennym: They were casted before, so I left them casted, but you're right.

joefitter: https://github.com/UKHomeOffice/asl-pages/pull/716/commits/9bc423de09d3137f837d4abb8e73e9f537c33a21#diff-2c902969d60f43f430908e85a50cf87fL26 WERE THEY THO

joefitter: Jesus Lenny.

lennym: LITERALLY YES!

Screenshot 2020-02-27 at 16 34 48

PR #727, opened 2020-03-04
Allow update of licence number for project stubs

wheelsandcogs:


pages/project/read/views/index.jsx
```diff
@@ -372,7 +372,16 @@ export default function ProjectLandingPage() {
 
         <Fragment>
           <dt><Snippet>fields.licenceNumber.label</Snippet></dt>
-          <dd>{model.licenceNumber ? model.licenceNumber : '-'}</dd>
+          <dd>
+            {model.licenceNumber ? model.licenceNumber : '-'}
+            {
+              model.isLegacyStub && allowedActions.includes('project.updateLicenceNumber') &&
```
lennym: This is ok without a feature flag because `isLegacyStub` will always be false, right?

wheelsandcogs: Yeah, it's impossible to have a legacy stub without the previous feature being enabled to start the process.

PR #728, opened 2020-03-05
Feature - PPL list updates

joefitter:
* Updated columns to active, inactive and drafts
* Added status col to inactive licences (expired, revoked and transferred)
* removed redundant sequelize filter

### DIFF
--- a/pages/project/content/index.js
+++ b/pages/project/content/index.js
@@ -55,8 +55,10 @@ module.exports = {
   },
   status: {
     active: 'Active',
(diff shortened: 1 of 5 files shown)


joefitter:
WIP until https://github.com/UKHomeOffice/asl-schema/pull/270

lennym:
This is only necessary or useful once there are transferred PPLs in the wild, right?

That is, can we hold off merging it until everything else has shipped?

joefitter:
yep 👍

PR #735, opened 2020-03-10
Show 5 year review date on PIL pdf header

wheelsandcogs:


pages/common/views/pdf/header.jsx
```diff
@@ -36,6 +36,10 @@ const Header = ({ store, model, licenceType, nonce, version }) => (
             <Fragment> | Amended: <strong>{format(model.amendedDate)}</strong></Fragment>
           }
 
+          { licenceType === 'pil' && model.status === 'active' &&
+            <Fragment> | 5 year review: <strong>{format(model.reviewDate)}</strong></Fragment>
```
lennym: Might be worth checking if this is always set. There was some work I did where if it's not set then it defaults to "updated at + 5 years" but I don't remember _where_ that was done.

wheelsandcogs: Should be fine, it's in public-api:
https://github.com/UKHomeOffice/asl-public-api/blob/master/lib/routers/profile/pil.js#L112

PR #737, opened 2020-03-10
Feature - PPL transfer

joefitter:
* pass user establishments and ppl transfer perms through to asl-projects
* removed transfer interface
* added success messages for transfer journey
* added logic to lookup messages from selected status -> actual status


pages/common/routers/success.js
```diff
@@ -24,22 +24,23 @@ module.exports = ({
   model = 'model',
   licence,
   type,
-  status
+  status,
+  refModel
```
lennym: What is a `refModel`?

joefitter: the model we care about - the one which has openTasks etc - usually req.model, but not in the case of projects - req.model === version, but we care about the project

lennym:
Raised #739 to deal with some of my 🐄 in and around this.

PR #742, opened 2020-03-11
Render markdown inside the activity log comments

wheelsandcogs:


pages/task/read/views/components/activity-log.jsx
```diff
@@ -68,7 +68,7 @@ const LogItem = ({ log, task }) => {
       {
         task.data.model === 'project' && <ExtraProjectMeta item={log} task={task} />
       }
-      <p className="comment">{log.comment}</p>
+      <Markdown className="comment" links={true}>{log.comment}</Markdown>
```
lennym: What's the logic behind `links={true}`? I'd probably want to disable links here unless there's a compelling reason.

wheelsandcogs: I thought it'd be useful to have them - e.g. if an inspector wants to link to some online guidance.

lennym: They are disabled by default because inspectors were accidentally making links in ppl comments by using square brackets. The same would apply here.

PR #745, opened 2020-03-16
Feature - PPL transfer content

joefitter:
* Added missing referred to inspector success messages
* Removed activity log filtering logic - this is now handled in workflow
* Updated activity log to show status and action from task


joefitter:
this is a WIP until https://github.com/UKHomeOffice/asl-workflow/pull/241/files is merged

PR #757, opened 2020-03-26
Prevent multiple tasks being raised for the same person/role/establishment

wheelsandcogs:
~Waiting on https://github.com/UKHomeOffice/asl-public-api/pull/236~

### DIFF
--- a/pages/role/apply/views/index.jsx
+++ b/pages/role/apply/views/index.jsx
@@ -29,7 +29,7 @@ const formatters = {
 };
 
 const Page = ({ addRoleTasks, schema }) => {
-  if (schema.type.options < 1) {
+  if (schema.type.options.length === 0) {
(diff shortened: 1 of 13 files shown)


pages/role/apply/views/index.jsx
```diff
@@ -27,11 +28,24 @@ const formatters = {
   }
 };
 
-const Page = () => {
+const Page = ({ addRoleTasks, schema }) => {
+  if (schema.type.options < 1) {
```
lennym: This feels like it's missing a `.length`

lennym: Also, if we mean `=== 0` can we say `=== 0`?

wheelsandcogs: sorted.

PR #759, opened 2020-03-27
Warn user that removing nacwo role will remove place relations

wheelsandcogs:


pages/role/apply/index.js
```diff
@@ -57,6 +57,7 @@ module.exports = settings => {
       };
 
       res.locals.static.addRoleTasks = addRoleTasks;
+      req.model.openTasks = []; // hide the open tasks warning on role forms as it is not applicable
```
lennym: I'm confused. Didn't we just do a load of work to add this?

lennym: Ahhh, I see the previous line now.

#homerhedge

PR #765, opened 2020-04-01
Add generic class for positioning csv downloads near tables

lennym:
Binding this style onto the tab content doesn't make sense when we have download links not in the context of tabs.

Add a generic "table-wrapper" class for positioning CSV download links near tables.

### DIFF
--- a/pages/establishment/licence-fees/views/index.jsx
+++ b/pages/establishment/licence-fees/views/index.jsx
@@ -91,7 +91,7 @@ export default function Fees({ tab, tabs, children, subtitle = '' }) {
             tabs.map(({ page, key, ...props }, index) => <Link key={index} page={page} label={<Snippet>{key}</Snippet>} {...props} />)
           }
         </Tabs>
-        <div className="tab-content">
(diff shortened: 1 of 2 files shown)


pages/establishment/licence-fees/views/index.jsx
```diff
@@ -91,7 +91,7 @@ export default function Fees({ tab, tabs, children, subtitle = '' }) {
             tabs.map(({ page, key, ...props }, index) => <Link key={index} page={page} label={<Snippet>{key}</Snippet>} {...props} />)
           }
         </Tabs>
-        <div className="tab-content">
```
wheelsandcogs, line 94: is there other `tab-content` styling? Do we still need the class if not?

joefitter: the class is still there

wheelsandcogs: That's why I'm asking, if we've removed the styling do we still need the class name? It's not doing any harm I guess.

PR #769, opened 2020-04-06
Restructure profile amendments content

lennym:
Task pages lookup content for comments depending on task type, so the content needs to be scoepd under "amendment" (which is the only type of profile task)

### DIFF
--- a/pages/task/read/content/profile.js
+++ b/pages/task/read/content/profile.js
@@ -7,7 +7,9 @@ module.exports = merge({}, profile, {
   },
   'sticky-nav': {
     changes: 'Amendment requested',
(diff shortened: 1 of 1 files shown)


joefitter:
sorrybox

pages/task/read/content/profile.js
```diff
@@ -7,7 +7,9 @@ module.exports = merge({}, profile, {
   },
   'sticky-nav': {
     changes: 'Amendment requested',
-    comments: 'Why are you making this amendment?'
+    comments: {
+      amendment:'Why are you making this amendment?'
```
joefitter, line 11:
```suggestion
      amendment: 'Why are you making this amendment?'
```

PR #770, opened 2020-04-09
NTS PDFs

wheelsandcogs:


pages/project-version/pdf/views/nts/components/duration.jsx
```diff
@@ -0,0 +1,11 @@
...
+
+export default function Duration({ version }) {
+  const duration = version.duration;
+  if (!duration) {
+    return;
+  }
+  let yearsLabel = duration.years === 1 ? 'Year' : 'Years';
+  let monthsLabel = duration.months === 1 ? 'Month' : 'Months';
+  return <p>{`${duration.years} ${yearsLabel} ${duration.months} ${monthsLabel}`}</p>;
+}
```
joefitter, line 11: do we not have standard renderers for fields like this in projects?

wheelsandcogs: There's this but it's a DL:
https://github.com/UKHomeOffice/asl-projects/blob/master/client/components/review-field.js#L37-L59

wheelsandcogs: and there's https://github.com/UKHomeOffice/asl-projects/blob/master/client/components/duration.js but I think that's an input?

joefitter: I would argue against displaying the same thing in a different way here - it's always a DL in granted, readonly, and PDF

pages/project-version/pdf/views/nts/components/retrospective-assessment.jsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import RAContent from '@asl/projects/client/constants/retrospective-assessment';
+import raApplies from '@asl/projects/client/helpers/retrospective-assessment';
```
lennym, line 3: This was refactored away into the API data. It's probably not worth importing the module for the single line of code it contains.

https://github.com/UKHomeOffice/asl-projects/blob/master/client/helpers/retrospective-assessment.js#L2

pages/project-version/pdf/views/nts/components/species-table.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { uniq, concat } from 'lodash';
+
+const groupSpeciesDetails = version => {
+  return (version.protocols || []).reduce((species, protocol) => {
+    if (protocol && protocol.speciesDetails && protocol.speciesDetails.length > 0) {
+      protocol.speciesDetails.map(details => {
+        const existingSpecies = species.find(s => s.value === details.value);
```
lennym, line 8: Using an object keyed by the value would be a simpler and more efficient way of collecting these.

It's probably doesn't matter on the scales we're talking about though.

pages/project-version/routes.js
```diff
@@ -23,6 +23,10 @@ module.exports = {
     path: '/pdf',
     router: pdf
   },
+  ntsPdf: {
+    path: '/pdf/nts',
+    router: pdf
+  },
```
lennym, line 29: I think this is completely redundant on the basis that `/pdf/nts` will have already been matched to `/pdf` and routed to the `pdf` router in the block above.

It's not doing any harm though.

wheelsandcogs: It's useful for the link here: https://github.com/UKHomeOffice/asl-pages/pull/770/files#diff-e44a7db1107d243b630f6e677a105d10R54 rather than passing a suffix.

lennym: Makes sense. 👍

PR #771, opened 2020-04-09
Feature - manage project collaborators

joefitter:
* added collaborators table
* added add users interface
* added router for removing users

### DIFF
--- a/pages/project/add-user/content/index.js
+++ b/pages/project/add-user/content/index.js
@@ -1,0 +1,15 @@
+const { merge } = require('lodash');
+const baseContent = require('../../content');
+
+module.exports = merge({}, baseContent, {
+  title: 'Grant access',
+  fields: {
+    profile: {
+      label: 'Who needs to be able to view this project?',
+      hint: 'If the person you’re looking for doesn’t appear in the list, ask a [HOLC at your establishment]({{profilesHolcFiltered}}) to give them access on your behalf.'
+    }
+  },
+  notifications: {
+    success: 'Access granted. The user has been notified.'
+  }
+});
--- a/pages/project/add-user/content/index.js
+++ b/pages/project/add-user/content/index.js
@@ -10,6 +10,6 @@ module.exports = merge({}, baseContent, {
     }
   },
   notifications: {
-    success: 'Access granted. The user has been notified.'
+    success: 'Access granted. Your colleague has been notified.'
   }
 });
--- a/pages/project/add-user/index.js
+++ b/pages/project/add-user/index.js
@@ -1,0 +1,50 @@
+const { stringify } = require('qs');
+const { page } = require('@asl/service/ui');
+const { form } = require('../../common/routers');
+const getSchema = require('./schema');
+
+module.exports = () => {
+  const app = page({ root: __dirname });
+
+  app.use((req, res, next) => {
+    req.model = {
+      id: `${req.project.id}-add-user`
+    };
+    const query = {
+      filters: {
+        roles: ['holc']
+      }
+    };
+    res.locals.static.profilesHolcFiltered = `${req.buildRoute('profile.list')}?${stringify(query)}`;
+    next();
+  });
+
+  app.use(form({
+    configure: (req, res, next) => {
+      req.api(`/establishment/${req.establishmentId}/profiles`, { query: { limit: 'all' } })
+        .then(({ json: { data } }) => {
+          const profiles = data.filter(profile => !req.project.collaborators.map(p => p.id).includes(profile.id));
+          req.form.schema = getSchema(profiles);
+        })
+        .then(() => next())
+        .catch(next);
+    }
+  }));
+
+  app.post('/', (req, res, next) => {
+    const profileId = req.form.values.profile;
+    req.api(`/establishment/${req.establishmentId}/project/${req.projectId}/collaborators/${profileId}`, { method: 'POST' })
+      .then(() => next())
+      .catch(next);
+  });
+
+  app.post('/', (req, res, next) => {
+    const id = req.model.id;
+    delete req.session.form[id];
+
+    req.notification({ key: 'success' });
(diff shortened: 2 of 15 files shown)


pages/project/add-user/index.js
```diff
@@ -0,0 +1,50 @@
...
+      }
+    };
+    res.locals.static.profilesHolcFiltered = `${req.buildRoute('profile.list')}?${stringify(query)}`;
+    next();
+  });
+
+  app.use(form({
+    configure: (req, res, next) => {
+      req.api(`/establishment/${req.establishmentId}/profiles`, { query: { limit: 'all' } })
+        .then(({ json: { data } }) => {
```
wheelsandcogs, line 25: is data always an array?

lennym: If the API has returned a 200 then yes, it should be. 

At least, if it's not then it's ok for things to loudly blow up I think.

pages/project/add-user/content/index.js
```diff
@@ -0,0 +1,15 @@
+const { merge } = require('lodash');
+const baseContent = require('../../content');
+
+module.exports = merge({}, baseContent, {
+  title: 'Grant access',
+  fields: {
+    profile: {
+      label: 'Who needs to be able to view this project?',
+      hint: 'If the person you’re looking for doesn’t appear in the list, ask a [HOLC at your establishment]({{profilesHolcFiltered}}) to give them access on your behalf.'
```
lennym, line 9: The link is a nice touch.

lennym: Except that it should probably be "admin" rather than HOLC, no?

pages/project/components/collaborators.jsx
```diff
@@ -0,0 +1,55 @@
+import React, { Fragment, useRef } from 'react';
+import { useSelector } from 'react-redux';
+import { Snippet } from '@asl/components';
+
+function RemoveUser({ id }) {
+  const removeUserUrl = useSelector(state => state.static.removeUserUrl);
+  const url = useSelector(state => state.static.url);
+  const formRef = useRef(null);
```
lennym, line 8: What's this for?

lennym: Oh. It's for hacking a link to submit the form.

Why not use a button? I think there are even styles for making buttons look like links.

Guess: `className="link"`

pages/project/content/index.js
```diff
@@ -63,5 +63,14 @@ module.exports = {
     transferred: 'Transferred',
     expired: 'Expired',
     revoked: 'Revoked'
+  },
+  collaborators: {
+    title: 'Colleagues granted access',
+    fields: {
+      name: 'Name',
+      email: 'Email',
+      action: ''
```
lennym, line 72: ?

pages/project/remove-user/index.js
```diff
@@ -0,0 +1,23 @@
...
+    req.api(`/establishment/${req.establishmentId}/project/${req.project.id}/collaborators/${profileId}`, { method: 'DELETE' })
+      .then(() => next())
+      .catch(next);
+  });
+
+  app.post('/', (req, res, next) => {
+    res.locals.static.content = merge(res.locals.static.content, content);
+    req.notification({ key: 'success' });
+    res.redirect(req.query.referrer);
+  });
```
lennym, line 20: Can we add a `GET` to this router that redirects to either the project page or edit collaborators page?

The reason being that if an error is thrown on the POST then there's a link to "try again" which reloads the same url on a GET. That should do _something_ other than 404.

pages/project/add-user/content/index.js
joefitter: I've asked the question - but you are probably right

pages/project/components/collaborators.jsx
joefitter: hahaha - that would be too easy! you get a black link, and when you hover it turns into a button.

pages/project/content/index.js
joefitter: yeah, in the content doc we have removed the heading for actions

pages/project/remove-user/index.js
joefitter: aah good plan

pages/project/add-user/index.js
joefitter: you'll always get at least yourself :)

pages/project/components/collaborators.jsx
lennym: We manage it ok for training modules. 

https://github.com/UKHomeOffice/asl-pages/blob/master/pages/pil/dashboard/views/section-details.jsx#L56

joefitter: ah - I think it's a problem if you use a `Button` not a `button`

lennym: You have to say it quietly or the CSS gets scared.

joefitter: 😆

lennym:
General question: what happens if you add a collaborator, but then they are removed from the establishment?

Answer is _probably_ (based on my understanding) that they retain access to the project, which is probably something we want to fix before this ships to prod.

pages/project/read/views/components/manage-access.jsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <Section
+      title={<Snippet>manageAccess.title</Snippet>}
+      content={<Snippet>manageAccess.content</Snippet>}
+    >
+      <p>
+        <Link page="project.addUser" label={<Snippet>manageAccess.action</Snippet>} />
+      </p>
+      {
+        !!model.collaborators.length && <Collaborators collaborators={model.collaborators} />
```
lennym, line 22: I'd be tempted to handle this in the component, but there's not necessarily any harm in having it here too.

PR #772, opened 2020-04-14
added collab name to success msg

joefitter:


lennym:
Can you do the removal one too pls?

joefitter:
there isn't one for removal - they don't get a notification

joefitter:
or are you suggesting just having "blah has been removed."

lennym:
It showed _something_ in S&T, and I remember seeing it and thinking "that's a bad message, it should have the name in it"

joefitter:
sure thing, I'll get right on that (don't tell Nadia) 😆

PR #774, opened 2020-04-15
Extract PDF renderer helper

lennym:
Extract a common PDF renderer component and reduce the amount of repeated rendering code used across page routers.

Adds a `noRender` option to all PDFs. This will prevent the PDF renderer assuming we're passing a mustache template and trying to check over a several megabyte string for a mustache token regex.

### DIFF
--- a/pages/establishment/pdf/index.jsx
+++ b/pages/establishment/pdf/index.jsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { createStore } from 'redux';
-import fetch from 'r2';
 import { Router } from 'express';
 import { renderToStaticMarkup } from 'react-dom/server';
 import groupBy from 'lodash/groupBy';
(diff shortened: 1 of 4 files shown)


pages/establishment/pdf/index.jsx
```diff
@@ -43,39 +44,16 @@ module.exports = settings => {
...
-          .response
+        pdf({ body, header, footer, hasStatusBanner })
           .then(response => {
-            if (response.status < 300) {
-              res.attachment(`${establishment.name}.pdf`);
-              response.body.pipe(res);
-            } else {
-              throw new Error(`Error generating PDF - generator responded ${response.status}`);
-            }
+            res.attachment(`${establishment.name}.pdf`);
```
wheelsandcogs, line 55: we probably want to run the establishment name through filenamify as well just incase.

lennym: That'll teach me to make whitespace changes...

PR #779, opened 2020-04-17
Added missing log content for 'updated' - this should not show unless…

joefitter:
… a task gets into an error state

### DIFF
--- a/pages/task/content/status.js
+++ b/pages/task/content/status.js
@@ -6,7 +6,8 @@ module.exports = {
     state: 'Case opened'
   },
   update: {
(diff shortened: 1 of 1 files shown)


pages/task/content/status.js
```diff
@@ -6,7 +6,8 @@ module.exports = {
     state: 'Case opened'
   },
   update: {
-    state: 'Case updated'
+    state: 'Case updated',
+    log: 'Case updated'
```
lennym, line 10:
```suggestion
    log: 'Updated by'
```

PR #781, opened 2020-04-17
Bugfix - missing PIL success messages

joefitter:
* Added success message for autoresolving pil application (LO revokes and resubmits)
* Added success message for non-licensing internal user submitted a revoked pil to licensing


lennym:
Not meaning to be _that guy_, but I think the first case should probably be a `resolved` rather than an `autoresolved` - i.e. it should leave a visible task behind.

joefitter:
hmm yeah maybe

PR #785, opened 2020-04-21
Feature - download ppl

joefitter:
* Added ppl download link and router

### DIFF
--- a/pages/project-version/ppl/index.js
+++ b/pages/project-version/ppl/index.js
@@ -1,0 +1,13 @@
+const { Router } = require('express');
+
+module.exports = () => {
+  const app = Router();
+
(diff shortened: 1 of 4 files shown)


pages/project-version/ppl/index.js
```diff
@@ -0,0 +1,13 @@
+const { Router } = require('express');
+
+module.exports = () => {
+  const app = Router();
+
+  app.get('/', (req, res, next) => {
+    const values = req.version.data || {};
+    res.attachment(`${values.title || 'Untitled project'}.ppl`);
```
lennym, line 8: `filenameify` pls

wheelsandcogs, line 8: might want to run this through `filenamify()`.

PR #791, opened 2020-04-30
Bugfix - dont try and render invalid date

joefitter:
* If expiry is missing dont try and render it


pages/task/list/views/table.jsx
```diff
@@ -54,7 +55,7 @@ const formatters = {
           }
           {
             continuation && (
-              firstExpiry
+              firstExpiry && isValid(firstExpiry)
```
lennym, line 58: It's probably worth rolling this into the `firstExpiry` bit above in case there's one with a valid date and another without.

PR #796, opened 2020-05-04
Continuation changes

joefitter:
* Replaced deadline component with text label
* Playback original contnuation RTE on task page if present

### DIFF
--- a/pages/task/read/content/project.js
+++ b/pages/task/read/content/project.js
@@ -10,7 +10,9 @@ module.exports = {
   continuation: {
     title: 'This is a project continuation',
     licence: 'From the licence',
-    expiry: 'Expiry'
+    expiry: 'Expiry',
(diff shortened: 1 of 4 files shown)


pages/task/read/content/project.js
```diff
@@ -10,7 +10,9 @@ module.exports = {
   continuation: {
     title: 'This is a project continuation',
     licence: 'From the licence',
-    expiry: 'Expiry'
+    expiry: 'Expiry',
+    rte: 'Original input',
+    label: 'Please state the licence number and expiry date of all these licences.'
```
lennym, line 15: Thanks for not trying to make this clever.

joefitter: it would have to have been clever as this question doesn't exist anymore

PR #804, opened 2020-05-11
Assign multiple NACWOs

wheelsandcogs:
~Needs 
https://github.com/UKHomeOffice/asl-components/pull/130
https://github.com/UKHomeOffice/asl-public-api/pull/248~

### DIFF
--- a/pages/common/assets/sass/style.scss
+++ b/pages/common/assets/sass/style.scss
@@ -393,7 +393,7 @@ details {
 }
 
 .govuk-table {
-  .nacwos p {
+  p.no-margins {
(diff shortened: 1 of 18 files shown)


pages/common/assets/sass/style.scss
```diff
@@ -391,3 +391,9 @@ details {
     padding-left: $govuk-gutter-half;
   }
 }
+
+.govuk-table {
+  .nacwos p {
```
lennym, line 396: Can we do something less _incredibly_ specific for the fairly common case where we want lists of things inside table cells?

PR #810, opened 2020-05-15
fix create/delete/amend place

wheelsandcogs:


pages/place/routers/confirm.js
```diff
@@ -59,7 +61,16 @@ module.exports = settings => {
...
 
-      next();
+      return next();
+    },
+    process: (req, res, next) => {
+      if (settings.method !== 'DELETE') {
+        req.session.form[req.model.id].values.roles = concat(
+          req.session.form[req.model.id].values.nacwos,
+          req.session.form[req.model.id].values.nvssqps
+        ).filter(Boolean);
```
lennym, line 71: Can we use `get`/`set` here to work better if the session expires?

pages/place/routers/confirm.js
```diff
@@ -48,6 +43,13 @@ module.exports = settings => {
         }
       });
 
+      const selectedNacwoIds = req.session.form[req.model.id].values.nacwos
+        ? req.session.form[req.model.id].values.nacwos.filter(Boolean)
+        : [];
+      const selectedNvsSqpIds = req.session.form[req.model.id].values.nvssqps
+        ? req.session.form[req.model.id].values.nvssqps.filter(Boolean)
+        : [];
```
lennym, line 51: Can we use `get` to not be dependent on the session being present please?

PR #815, opened 2020-05-26
Bugfix - completed task diffs

joefitter:
* Updated diff component labels when task completed to previous and changed to
* updated task read router to use modelData rather than looking up model
* Updated diff text and custom diff components to show correct labels when task is complete

### DIFF
--- a/pages/task/read/routers/read.js
+++ b/pages/task/read/routers/read.js
@@ -105,8 +105,8 @@ module.exports = () => {
     const action = req.task.data.action;
     const model = req.task.data.model;
     if (action === 'update' || action === 'delete' || action === 'update-conditions') {
-      // if task is resolved, get previous values from task
-      if (req.task.status === 'resolved') {
+      // if task is closed, get previous values from task
+      if (!req.task.isOpen) {
         res.locals.static.values = get(req.task, 'activityLog[0].event.data.modelData');
 
         if (model === 'establishment') {
--- a/pages/task/read/routers/read.js
+++ b/pages/task/read/routers/read.js
@@ -104,19 +104,22 @@ module.exports = () => {
   app.use((req, res, next) => {
     const action = req.task.data.action;
     const model = req.task.data.model;
-    if (model === 'establishment') {
-      res.locals.static.values = req.establishment;
-      return next();
-    }
     if (action === 'update' || action === 'delete' || action === 'update-conditions') {
-      if (model === 'profile' && req.user.profile.id === req.task.data.id) {
-        res.locals.static.values = req.user.profile;
-        return next();
-      }
-
       // if task is resolved, get previous values from task
       if (req.task.status === 'resolved') {
         res.locals.static.values = get(req.task, 'activityLog[0].event.data.modelData');
+
+        if (model === 'establishment') {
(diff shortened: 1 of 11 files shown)


joefitter:
WIP until workflow changes to modelData - also needs to ensure b/c for in flight tasks

pages/task/read/routers/read.js
```diff
@@ -104,11 +104,20 @@ module.exports = () => {
   app.use((req, res, next) => {
     const action = req.task.data.action;
     const model = req.task.data.model;
-    if (model === 'establishment') {
-      res.locals.static.values = req.establishment;
-      return next();
-    }
     if (action === 'update' || action === 'delete' || action === 'update-conditions') {
+      // if task is resolved, get previous values from task
+      if (req.task.status === 'resolved') {
```
lennym, line 109: This should check if the task is open rather than a specific status, otherwise the diffs will be wonky for non-resolved closed tasks.

I believe there's a decorator that sets a property like `isOpen` on the task.

pages/task/read/views/models/establishment.jsx
```diff
@@ -19,6 +19,8 @@ export default function Establishment({ task, values }) {
   const establishment = useSelector(state => state.static.establishment);
   const showNprc = establishment.nprc && (!establishment.pelh || establishment.pelh.id !== establishment.nprc.id);
 
+  const isComplete = task.status === 'resolved';
```
lennym, line 22: As above.

pages/task/read/views/models/place.jsx
```diff
@@ -29,6 +29,8 @@ export default function Playback({ task, schema, values, allowSubmit }) {
   const actionableNextSteps = task.nextSteps.filter(step => !nopes.includes(step.id));
   const canEditRestictions = isAsru && !!actionableNextSteps.length;
 
+  const isComplete = task.status === 'resolved';
```
lennym, line 32: Same

pages/task/read/views/models/project.jsx
```diff
@@ -17,19 +17,31 @@ import Deadline from '../components/deadline';
...
 import { schema as projectSchema } from '../../../../project/schema';
 
-const selector = ({ static: { project, establishment, version } }) => ({ project, establishment, version });
+const selector = ({ static: { project, establishment, version, values } }) => ({ project, establishment, version, values });
 
 // declarations can be 'Yes', 'No', or 'Not yet'
 const declarationConfirmed = declaration => declaration && declaration.toLowerCase() === 'yes';
 
-function EstablishmentDiff({ task }) {
+function EstablishmentDiff({ task, isComplete }) {
```
lennym, line 25: Completeness can be trivially determined from the task, and is entirely linked to the task, so I wouldn't pass it as a separate prop.

lennym:
I think you want to replace all instances of the `const isComplete = task.status === 'resolved'` with an inspection of `task.isOpen`.

pages/task/read/routers/read.js
joefitter: if they are non-resolved then the data should be looked up from the model, it's only resolved/autoresolved we want, but autoresolved won't be in the task list

PR #829, opened 2020-06-11
Sort comments newest first

wheelsandcogs:
~waiting on https://github.com/UKHomeOffice/asl-projects/pull/578~


lennym:
Happy to merge this as is, since it's exactly as fragile as it was before, but I can't help thinking it would be better to explicitly sort them if there is an expectation that they're ordered in a particular way.

PR #832, opened 2020-06-12
Feature - global profile

joefitter:
* Added global profile page from internal ui
* Stripped out asru actions
* allow own profile and asru profiles
* show edit links if own profile
* show related tasks if not asru user

### DIFF
--- a/pages/global-profile/components/asru-roles.jsx
+++ b/pages/global-profile/components/asru-roles.jsx
@@ -1,0 +1,29 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { Snippet } from '@asl/components';
+
+const roles = ['asruAdmin', 'asruLicensing', 'asruInspector'];
+
+class Roles extends React.Component {
+
+  render () {
+    const { model } = this.props;
+    const userRoles = roles.filter(role => model[role]);
+
+    return (
+      <div className="asru-roles">
+        <h3><Snippet>asru.roles.title</Snippet></h3>
+        <ul>
+          {
+            userRoles.map(role => <li key={ role }><Snippet>{ `asru.roles.${role}` }</Snippet></li>)
+          }
+        </ul>
+      </div>
+    );
+  }
+
+}
+
+const mapStateToProps = ({ model }) => ({ model });
+
+export default connect(mapStateToProps)(Roles);
--- a/pages/global-profile/components/asru-roles.jsx
+++ b/pages/global-profile/components/asru-roles.jsx
@@ -1,29 +1,19 @@
 import React from 'react';
-import { connect } from 'react-redux';
+import { useSelector } from 'react-redux';
 import { Snippet } from '@asl/components';
 
-const roles = ['asruAdmin', 'asruLicensing', 'asruInspector'];
+export default function Roles({ roles = ['asruLicensing', 'asruInspector'] }) {
+  const model = useSelector(state => state.model);
+  const userRoles = roles.filter(role => model[role]);
 
-class Roles extends React.Component {
-
-  render () {
-    const { model } = this.props;
-    const userRoles = roles.filter(role => model[role]);
-
-    return (
-      <div className="asru-roles">
-        <h3><Snippet>asru.roles.title</Snippet></h3>
-        <ul>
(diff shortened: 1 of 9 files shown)


pages/global-profile/views/index.jsx
```diff
@@ -0,0 +1,135 @@
...
+              model.certificates && model.certificates.length > 0
+                ? <Modules certificates={model.certificates} />
+                : <p><em><Snippet>pil.training.none</Snippet></em></p>
+            }
+          </Fragment>
+        )
+      }
+
+      {
+        model.asruUser && <AsruRoles />
```
lennym, line 106: This feels like it's leaking way too much information to external users. Shouldn't this be only for ASRU users viewing the profile?

joefitter: not according to the ticket - this is just a list of asru roles, so if the profile you are looking at is an asru user it will say whether they are an inspector/LO

lennym: I think it's fine to expose if they are an inspector or LO, but I'd be wary of exposing whether they have any other roles.

Can we limit the output to _only_ inspector/LO?

joefitter: ah, yeah good shout

pages/global-profile/components/asru-roles.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { Snippet } from '@asl/components';
+
+const roles = ['asruAdmin', 'asruLicensing', 'asruInspector'];
```
lennym, line 5:
```suggestion
const roles = ['asruLicensing', 'asruInspector'];
```

pages/global-profile/views/index.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+class Index extends React.Component {
+
+  render () {
+
+    const model = this.props.model;
+    const isOwnProfile = this.props.isOwnProfile;
+    const hasEstablishments = !!(model.establishments || []).length;
+
+    const showEmail = this.props.asruUser || model.asruLicensing || this.props.isOwnProfile;
```
lennym, line 24: Is there a reason we show emails for licensing officers but not inspectors?

(sorry if this is in the ticket)

joefitter: yes - it's in the ticket

lennym: Wild!

pages/global-profile/views/index.jsx
```diff
@@ -11,22 +11,46 @@ import Modules from '../../profile/read/views/modules';
...
-  render () {
+export default function Index({ dedupe, AsruRolesComponent, children }) {
+  const { model, allowedActions, asruUser, isOwnProfile } = useSelector(selector, shallowEqual);
+  const hasEstablishments = !!(model.establishments || []).length;
 
-    const model = this.props.model;
-    const isOwnProfile = this.props.isOwnProfile;
-    const hasEstablishments = !!(model.establishments || []).length;
+  const showEmail = asruUser || model.asruLicensing || isOwnProfile;
+  const showDob = (asruUser || isOwnProfile) && model.dob;
```
lennym, line 35: I think we probably need to filter this on the server-side and then simply render here if it is defined.

Otherwise the data will be accessible through devtools.

pages/global-profile/content/index.js
```diff
@@ -0,0 +1,26 @@
...
+const profileContent = require('../../profile/read/content');
+const pilContent = require('../../task/read/content/pil');
+const tasklistContent = require('../../task/list/content');
+
+module.exports = merge({}, pilContent, profileContent, tasklistContent, {
+  establishment: {
+    link: 'About this establishment'
+  },
+  notifications: {
+    success: 'Changes saved!'
```
wheelsandcogs, line 11: Tiny thing - we could just fall back on the default here: 
https://github.com/UKHomeOffice/asl-pages/blob/master/pages/common/content/index.js#L178-L180

or at least keep them consistent (`!` vs `.`)

PR #837, opened 2020-06-16
Add alternate phone number field to profiles

wheelsandcogs:


lennym:
Is there a reason we alternate between "Alternate" and "Alt."?

I'd suggest prevailing good practice would be to not abbreviate unless absolutely necessary.

PR #838, opened 2020-06-16
Add help message for basic users on project list

wheelsandcogs:


pages/project/list/index.js
```diff
@@ -29,7 +29,10 @@ module.exports = settings => {
       next();
     },
     locals: (req, res, next) => {
+      const establishment = req.user.profile.establishments.find(e => e.id === req.establishmentId);
       set(res.locals, 'static.status', req.query.status || 'active');
+      set(res.locals, 'static.isBasicUser', !!(establishment && establishment.role === 'basic'));
```
lennym, line 34: Could we tie this to `project.read.all` permissions rather than the permission level, just to make it _a little_ more resilient to any future config changes?

PR #845, opened 2020-06-24
Feature - remove inspector declaration for first submission

joefitter:
* only show the inspector declaration if the project has an existing task

### DIFF
--- a/pages/project-version/update/submit/index.js
+++ b/pages/project-version/update/submit/index.js
@@ -40,7 +40,8 @@ module.exports = settings => {
   app.use(
     form({
       configure: (req, res, next) => {
-        const schema = getSchema(req.version.type, req.user.profile.asruUser);
+        const hasOpenTask = !!get(req.project, 'openTasks', []).length;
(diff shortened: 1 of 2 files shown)


joefitter:
This will likely break most of the tests so marking as WIP until a test fix is ready to go

pages/project-version/update/submit/index.js
```diff
@@ -40,7 +40,8 @@ module.exports = settings => {
   app.use(
     form({
       configure: (req, res, next) => {
-        const schema = getSchema(req.version.type, req.user.profile.asruUser);
+        const hasAuthority = get(req.project, 'openTasks[0].data.meta.authority') === 'Yes';
```
lennym, line 43:
```suggestion
        // if application has previously been approved then this is a resubmission and we can show the inspector ready question
        const hasAuthority = get(req.project, 'openTasks[0].data.meta.authority') === 'Yes';
```

lennym: Because it's a hack.

joefitter: heh

PR #848, opened 2020-06-25
Increase task list row count to 20 for ASRU users

lennym:


pages/task/list/router.js
```diff
@@ -16,6 +16,10 @@ module.exports = ({
   apiPath = '/tasks',
   schema = defaultSchema
 } = {}) => datatable({
+  configure: (req, res, next) => {
+    req.query.rows = req.query.rows || req.user.profile.asruUser ? 20 : 10;
```
wheelsandcogs, line 20:
```suggestion
    req.query.rows = req.query.rows || (req.user.profile.asruUser ? 20 : 10);
```
not necessary but makes it a bit clearer

lennym: 👍

PR #850, opened 2020-06-30
Feature - PPL submit declaration

joefitter:
If admin user submits a ppl then show a declaration, and remove warning text about requiring endorsement

### DIFF
--- a/pages/project-version/update/submit/index.js
+++ b/pages/project-version/update/submit/index.js
@@ -15,6 +15,19 @@ module.exports = settings => {
   });
 
   app.use((req, res, next) => {
+    const params = {
+      projectId: req.project.id,
+      establishmentId: req.establishmentId
+    };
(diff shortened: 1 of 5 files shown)


pages/project-version/update/submit/index.js
```diff
@@ -14,6 +14,19 @@ module.exports = settings => {
...
   });
 
+  app.use((req, res, next) => {
+    const params = {
+      projectId: req.project.id,
+      establishmentId: req.establishmentId
+    };
+    req.user.can('project.endorse', params)
+      .then(canEndorse => {
+        res.locals.static.canEndorse = canEndorse;
```
lennym, line 24: Is this not already accessible from `allowedActions`?

joefitter: ooh probably

PR #861, opened 2020-07-08
removed all page breaks from project PDFs apart from img

joefitter:


pages/common/assets/sass/pdf/index.scss
```diff
@@ -122,27 +112,10 @@ dl dd {
   text-indent: 5px;
 }
 
-.break {
-  page-break-after: always;
-}
```
lennym, line 127: This one probably wants to stay.

joefitter: YTHO?

joefitter: if we don't want to throwing breaks around then surely this is doing exactly that?

lennym:
I think this goes a bit too far. We probably want to get rid of all of the `avoid` but keep the `always`.

pages/common/assets/sass/pdf/index.scss
lennym: Because if there's an element with a classname of `.break` it implies that the page break is deliberate.

If it shouldn't have a page break then don't put the `.break` element in.

PR #868, opened 2020-07-14
Feature - Added RA required labels to PPL list

joefitter:
* Show status labels on each tab in ppl list
* Show RA Required label if applic


joefitter:
This will be a WIP until we are confident the RA migration has run successfully

PR #871, opened 2020-07-15
Add route for transferring ppl draft to another est

wheelsandcogs:


pages/project/read/views/index.jsx
```diff
@@ -63,7 +63,7 @@ function CurrentVersion({ model }) {
 function OpenTask({ model }) {
   const { openTask, editable, canUpdate, asruUser } = useSelector(state => state.static);
 
-  if (!openTask) {
+  if (!openTask || !openTask.data) {
```
lennym, line 66: What case is this catching?

wheelsandcogs: If workflow is down, `openTask` is an object that looks like `{ action: 'failed' }` so there's no `data` prop and the use of it below causes the page to error rather than just skipping the stuff that needs the tasks. It might never happen outside of local env when you forget to start workflow.

lennym: Makes sense. Thanks.

PR #876, opened 2020-07-23
Generic success pages

wheelsandcogs:
~Needs https://github.com/UKHomeOffice/asl-public-api/pull/266~


pages/pil/dashboard/index.js
```diff
@@ -22,7 +23,8 @@ module.exports = settings => {
       opts.json.data.establishment = req.model.establishment;
     }
 
-    return req.api(`/establishment/${req.establishmentId}/profiles/${req.profileId}/pil/${req.pilId}/${action}`, opts);
+    return req.api(`/establishment/${req.establishmentId}/profiles/${req.profileId}/pil/${req.pilId}/${action}`, opts)
+      .then(saveTaskIdToSession(req.session));
```
joefitter, line 27: this could do with moar `.catch` - I know it wasn't there before, but while you're in there :)

pages/pil/dashboard/index.js
```diff
@@ -104,17 +106,14 @@ module.exports = settings => {
 
   app.post('/', (req, res, next) => {
     sendData(req)
-      .then(() => delete req.session.form[req.model.id])
-      .then(() => res.redirect(req.buildRoute('pil.update', { suffix: 'success' })))
+      .then(() => {
+        delete req.session.form[req.model.id];
+        return res.redirect(req.buildRoute('pil.update', { suffix: 'success' }));
+      })
```
joefitter, line 112: does this need to do the task/session stuff too?

pages/common/helpers/index.js
```diff
@@ -16,8 +16,16 @@ const getNacwoById = (req, id) =>
     .then(nacwo => nacwo && nacwo.profile)
     .catch(err => Promise.reject(err));
 
+const saveTaskIdToSession = session => response => {
+  session.success = {
+    taskId: get(response, 'json.data.id')
+  };
+  return response;
+};
```
joefitter, line 24: I know _why_ you have done this, but it seems a little too side-effecty for a helper function, it relies on being called in a very specific context, I think it would feel nicer to have an implementation that looks more like this

```
.then(response => {
  saveTaskIdToSession(req, response);
  return response;
})
```

but then it probably doesn't warrant having a helper at all, maybe I need to care less

pages/pil/dashboard/index.js
wheelsandcogs: Anywhere we have a `sendData()` function I've added it there, because in some places we wrap `sendData()` in a call to `updateDataFromTask()` and I was having issues with saving the task id to the session in those cases.

pages/common/helpers/index.js
wheelsandcogs: originally there was no helper function and every router was saving to the session itself - but yeah if I simplify this any further (e.g. passing in the taskId instead of the response) it feels like I might as well just go back to not having a helper.

joefitter: I think the part that bothers me is the need to return the response, perhaps it would be better to have this as the last step in the promise chain, before catch, then it can just do it's thing without caring about what comes next?

joefitter: also "maybe I need to care less"

pages/common/content/success-messages.js
```diff
@@ -1,781 +1,112 @@
...
+  'revoked': {
+    panel: {
+      title: 'Revoked'
+    },
+    whatNext: {
+      body: `This licence has been revoked.`,
+      internal: 'The relevant people have been emailed.'
+    }
   }
 };
```
joefitter, line 112: 🎉

pages/common/helpers/index.js
joefitter: I realise that the "calling last" thing is not an option, because redirects, so please disregard that 🤷

pages/pil/dashboard/index.js
joefitter: carry on then 👍

pages/pil/dashboard/index.js
wheelsandcogs: This is called from the router below, which has a `catch()` - that should be good enough right?

joefitter: I didn't look that far down, as long as there's a catch somewhere in the promise chain then it shouldn't hang on error

pages/establishment/update/routers/confirm.js
```diff
@@ -22,7 +23,8 @@ module.exports = () => {
       }, params)
     };
 
-    return req.api(`/establishment/${req.establishmentId}`, opts);
+    return req.api(`/establishment/${req.establishmentId}`, opts)
+      .then(saveTaskIdToSession(req.session));
```
lennym, line 27: This one should move down to live alongside the other handling code at L67-ish so it is structured consistently with other implementations.

wheelsandcogs: It's here because of L32 - we wrap the `sendData()` function in `updateDataFromTask()` which does not return the original response from sendData

wheelsandcogs: Basically anywhere we've got a separate `sendData()` function, I've added it there because it's sometimes gets used with additional middleware.

PR #879, opened 2020-07-31
Feature - training in profiles

joefitter:
* Added new training in profiles section encompassing exemptions
* Removed training from pil - replace with page playing back the training table and confirmation it is up to date
* Added redirect to session, provides link back to pil/project
* Added training modules to project init


joefitter:
WIP until https://github.com/UKHomeOffice/asl-projects/pull/600

pages/pil/training/schema/index.js
```diff
@@ -1,6 +1,19 @@
...
+  update: {
+    inputType: 'radioGroup',
+    format: toBoolean,
+    options: [
+      {
+        label: 'Yes, I need to update it',
+        value: true
+      },
+      {
+        label: 'No, that looks fine',
```
lennym, line 13: image

pages/success/index.js
```diff
@@ -98,6 +98,7 @@ module.exports = () => {
 
   app.use((req, res, next) => {
     const taskId = get(req.session, 'success.taskId');
+    console.log('HERE', taskId);
```
lennym, line 101:
```suggestion
```

pages/training/modules/index.js
```diff
@@ -0,0 +1,41 @@
...
+        const isExemption = get(req.session, `form[${req.model.id}].values.isExemption`);
+        req.form.schema = getSchema(isExemption);
+        next();
+      },
+      locals(req, res, next) {
+        res.locals.static.isExemption = get(req.session, `form[${req.model.id}].values.isExemption`);
+        next();
+      }
+    })(req, res, next);
+  });
```
lennym, line 28: This is nested inside itself in a weird way. Just do `app.use(form({...}))` directly.

pages/success/index.js
joefitter: lol

pages/pil/training/schema/index.js
joefitter: I know right

pages/profile/read/views/index.jsx
```diff
@@ -50,6 +50,20 @@ const Index = ({
           <Profile profile={model} establishment={establishment} allowedActions={allowedActions} isOwnProfile={isOwnProfile} />
         </div>
       </div>
+      {
+        allowedActions.includes('training.read') && (
+          <section className="profile-section">
+            <h3><Snippet>training</Snippet></h3>
+            {
+              !!certificates.length && <TrainingSummary certificates={certificates} />
```
lennym, line 58: It feels like this check would be better done inside `<TrainingSummary />`

i.e.

```
if (!certificates || !certificates.length) {
  return null;
}
```

lennym: It would negate the need to default the certificates here as well.

joefitter: it already is, I just need to remove the check here

PR #882, opened 2020-08-03
add classes to lists for easy selection

wheelsandcogs:


pages/pil/procedures/views/diff.jsx
```diff
@@ -88,7 +88,7 @@ export default function ProceduresDiff({ before, after }) {
               </ul>
             </td>
             <td>
-              <ul>
+              <ul className="current">
```
lennym, line 91: Should this be:

```suggestion
              <ul className="proposed">
```

wheelsandcogs: Yep!

PR #885, opened 2020-08-03
Only show primary est name on PIL pdf to holding est users

wheelsandcogs:


pages/pil/pdf/index.jsx
```diff
@@ -20,6 +20,9 @@ module.exports = settings => {
       licenceHolder: req.profile
     };
 
+    const holdingEstablishmentId = req.pil.establishmentId;
+    const canReadHoldingEstablishment = !!req.user.profile.establishments.find(e => e.id === holdingEstablishmentId);
```
lennym, line 24: It feels a little like we have different versions of this logic spread around in a few different places. Specifically the logic used here is slightly different to the logic used in the web UI when viewing the same PIL. 

I'd probably look to streamline in the same way as the web UI has and infer from the presence of the establishment in the API response as to whether to render or not.

PR #887, opened 2020-08-05
better handling of non-amendment PIL tasks

wheelsandcogs:


lennym:
I don't think this is the right approach. Reviews aren't the special case that we need to _not_ show changes for, but rather amendments are the only case where we _do_ want to show changes, and everything else should simply show the current state.

I think implemented this way then revocations will have the same issue (although admittedly it's less wrong in that case).

I'd pass the task action into the summary/diff components and handle amendments as the special case.

PR #891, opened 2020-08-07
Move awerb question to admin ppl endorsement

wheelsandcogs:
waiting on 
~https://github.com/UKHomeOffice/asl-public-api/pull/272~
~https://github.com/UKHomeOffice/asl-workflow/pull/301~

### DIFF
--- a/pages/task/schema/confirm.js
+++ b/pages/task/schema/confirm.js
@@ -1,10 +1,21 @@
+const { get } = require('lodash');
+const { getAwerbQuestion } = require('../../project-version/update/submit/schema');
+
 module.exports = (task, chosenStatus) => {
   const commentRequired = stepId => {
     const nextStep = task.nextSteps.find(nextStep => nextStep.id === stepId);
     return nextStep && nextStep.commentRequired;
   };
 
-  return {
+  const awerbRequired = task => {
+    // awerb question is required for project applications / amendments that haven't yet been marked as awerbed
+    const model = get(task, 'data.model');
+    const status = get(task, 'status');
+    const wasAwerbed = get(task, 'data.meta.awerb') === 'Yes';
+    return model === 'project' && status === 'awaiting-endorsement' && !wasAwerbed;
+  };
+
+  let schema = {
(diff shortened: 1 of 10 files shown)


pages/task/schema/confirm.js
```diff
@@ -1,10 +1,21 @@
...
     return nextStep && nextStep.commentRequired;
   };
 
-  return {
+  const awerbRequired = task => {
+    // awerb question is required for project applications / amendments that haven't yet been marked as awerbed
+    const model = get(task, 'data.model');
+    const status = get(task, 'status');
+    const wasAwerbed = get(task, 'data.meta.awerb') === 'Yes';
+    return model === 'project' && status === 'awaiting-endorsement' && !wasAwerbed;
```
lennym, line 15: Why might a project task require endorsement but already have been awerbed?

pages/task/schema/confirm.js
```diff
@@ -14,4 +25,18 @@ module.exports = (task, chosenStatus) => {
...
     }
   };
+
+  const isAmendment = get(task, 'type') === 'amendment';
+
+  if (awerbRequired(task)) {
+    schema = {
+      awerb: {
+        ...getAwerbQuestion(isAmendment),
+        automapReveals: true
```
lennym, line 35: Is there a reason we want this here, but not in the other instances of the same thing?

wheelsandcogs: This is a new feature @joefitter added that means you don't need to manually map the reveals, e.g. https://github.com/UKHomeOffice/asl-pages/blob/master/pages/project-version/update/submit/views/index.jsx#L6-L26

lennym: How come it is _only_ used here and not in the other places we ask the question though?

pages/task/schema/confirm.js
wheelsandcogs: This schema is shared by all the task confirm pages, and we only want to show the Awerb question when a project is being endorsed.

PR #900, opened 2020-08-10
Evaluate conditional fields in change PPL holder form

lennym:
Get the active project version and use to evaluate which of the conditional fields in the experience section of the PPL scehma should be used in the change PPL holder journey.


lennym:
This is broken on the task view. Do not merge.

lennym:
Task view is fixed, but dependent on https://github.com/UKHomeOffice/asl-projects/pull/612

PR #912, opened 2020-08-18
inherited restrictions formatter was breaking the RestrictionsField c…

wheelsandcogs:
…omponent


wheelsandcogs:
Fixed in https://github.com/UKHomeOffice/asl-components/pull/154/files instead

PR #913, opened 2020-08-20
rename awaiting review to awaiting decision

wheelsandcogs:


pages/task/content/status.js
```diff
@@ -98,7 +98,7 @@ module.exports = {
     log: 'Submitted by'
   },
   'ntco-endorsed': {
-    state: 'Awaiting review',
+    state: 'Awaiting decision',
```
wheelsandcogs, line 101: Not sure where this is actually used, but changed for completeness.

pages/task/read/content/pil.js
```diff
@@ -28,7 +28,7 @@ module.exports = merge({}, baseContent, {
       }
     },
     endorsed: {
-      state: 'Awaiting review',
+      state: 'Awaiting decision',
```
wheelsandcogs, line 31: Same here, not sure where this is actually used

PR #914, opened 2020-08-20
Split out current task status into own section in the task view

wheelsandcogs:


pages/task/content/status.js
```diff
@@ -95,7 +100,8 @@ module.exports = {
   },
   'with-ntco': {
     state: 'Awaiting endorsement',
-    log: 'Submitted by'
+    log: 'Submitted by',
+    currentlyWith: 'Currently with: Named Training and Competency Officer'
```
lennym, line 104: Is this status _actually_ used for things that are with NTCO, or is it only there for b/c?

pages/task/content/status.js
```diff
@@ -109,12 +115,17 @@ module.exports = {
...
+    currentlyWith: 'Currently with: Home Office Licensing Officer'
   },
   // TODO: content
   'awaiting-endorsement': {
     state: 'Awaiting endorsement',
-    log: 'Submitted by'
+    log: 'Submitted by',
+    currentlyWith: {
+      pil: 'Currently with: Named Training and Competency Officer',
+      project: 'Currently with: Establishment Admin'
```
lennym, line 127: I think this answers my question.

pages/task/content/status.js
wheelsandcogs: think it's legacy.

pages/task/content/status.js
wheelsandcogs: 👍

PR #918, opened 2020-08-24
Activity log awerb/authority/ready question playback

wheelsandcogs:
~Will need rebasing once https://github.com/UKHomeOffice/asl-pages/pull/916 is merged, this is a continuation of that work.~

### DIFF
--- a/pages/common/assets/sass/style.scss
+++ b/pages/common/assets/sass/style.scss
@@ -132,6 +132,16 @@ dl.continuation {
       float: right;
     }
   }
+
+  .declarations {
+    p {
+      margin: 0;
+
+      &.question {
+        margin-top: $govuk-gutter-half;
+      }
+    }
+  }
 }
 
 .preserve-whitespace {
@@ -505,5 +515,13 @@ details {
 
   td {
     vertical-align: top;
+  }
+}
+
+#submitted-version {
+  .declarations {
+    .question span {
+      font-weight: bold;
+    }
   }
 }
--- a/pages/common/assets/sass/style.scss
+++ b/pages/common/assets/sass/style.scss
@@ -124,11 +124,6 @@ dl.continuation {
   }
 
   .log-item {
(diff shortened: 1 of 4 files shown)


pages/task/read/views/components/activity-log.jsx
```diff
@@ -1,43 +1,45 @@
...
 import classnames from 'classnames';
 import get from 'lodash/get';
-import { Link, Snippet, Markdown } from '@asl/components';
+import { Link, Snippet, Markdown, Inset } from '@asl/components';
 import { dateFormat } from '../../../../../constants';
 import format from 'date-fns/format';
+import PplDeclarations from './ppl-declarations';
 
-const getProfileLink = ({ id, name, establishmentId, asruUser }) => {
+function ProfileLink({ id, name, establishmentId, asruUser }) {
```
joefitter, line 9: I like this, if it looks like a component, make it a component 👍

pages/common/assets/sass/style.scss
```diff
@@ -512,3 +517,11 @@ details {
     vertical-align: top;
   }
 }
+
+#submitted-version {
+  .declarations {
+    .question span {
+      font-weight: bold;
```
joefitter, line 524: `<strong>` makes them bold without having to add extra styles

wheelsandcogs: We re-use the component in two places with different styling - in the activity log it's designed as non-bold and in the "Latest submission" section it's designed as bold, so this was the easiest way to get both.

pages/task/read/views/components/activity-log.jsx
lennym: I agree. Nice.

pages/common/assets/sass/style.scss
joefitter: seems legit

PR #922, opened 2020-08-27
the edit link on the delete place confirm page was broken

wheelsandcogs:
`settings.page` was undefined - just pass the actual route string instead.

### DIFF
--- a/pages/place/delete/index.js
+++ b/pages/place/delete/index.js
@@ -44,6 +44,7 @@ module.exports = settings => {
 
   app.use('/confirm', (req, res, next) => {
     confirm({
+      page: 'place.delete',
       method: 'DELETE',
       apiUrl: `/establishment/${req.establishmentId}/place/${req.model.id}`
     })(req, res, next);
--- a/pages/place/routers/confirm.js
+++ b/pages/place/routers/confirm.js
@@ -71,7 +71,7 @@ module.exports = settings => {
       if (req.session.form && req.session.form[req.model.id]) {
         return next();
       }
-      return res.redirect(req.buildRoute('place.delete'));
+      return res.redirect(req.buildRoute(settings.page));
     },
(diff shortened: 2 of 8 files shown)


pages/place/routers/confirm.js
```diff
@@ -71,11 +71,11 @@ module.exports = settings => {
       if (req.session.form && req.session.form[req.model.id]) {
         return next();
       }
-      return res.redirect(req.buildRoute(settings.page));
+      return res.redirect(req.buildRoute('place.delete'));
     },
     editAnswers: (req, res, next) => {
       delete req.session.form[req.model.id].validationErrors;
-      return res.redirect(req.buildRoute(settings.page));
+      return res.redirect(req.buildRoute('place.delete'));
```
lennym, line 78: This feels too specifically "delete" for a router which is not specifically "delete".

wheelsandcogs: Good point - will find a proper solution

pages/place/delete/index.js
```diff
@@ -44,6 +44,7 @@ module.exports = settings => {
 
   app.use('/confirm', (req, res, next) => {
     confirm({
+      page: 'place.delete',
```
wheelsandcogs, line 47: This is the bug fix, everything else is just tidying up issues I noticed while testing.

pages/place/routers/confirm.js
```diff
@@ -71,7 +71,7 @@ module.exports = settings => {
       if (req.session.form && req.session.form[req.model.id]) {
         return next();
       }
-      return res.redirect(req.buildRoute(settings.page));
+      return res.redirect(req.buildRoute('place.delete'));
```
lennym, line 74: This looks like it possibly needs reverting.

wheelsandcogs: was just testing you were awake 😄

PR #926, opened 2020-09-04
Add abbreviations to the named people acronyms

wheelsandcogs:
~Needs https://github.com/UKHomeOffice/asl-components/pull/163~

### DIFF
--- a/pages/place/formatters/index.jsx
+++ b/pages/place/formatters/index.jsx
@@ -1,15 +1,13 @@
 import React, { Fragment } from 'react';
-import { defineValue, joinAcronyms, labelFromCode } from '../../common/formatters';
+import { joinAcronyms, labelFromCode } from '../../common/formatters';
 
 export default {
   suitability: {
     format: joinAcronyms,
-    label: defineValue,
     mapOptions: labelFromCode
   },
   holding: {
     format: joinAcronyms,
-    label: defineValue,
     mapOptions: labelFromCode
   },
   restrictions: {
--- a/pages/place/list/content/index.js
+++ b/pages/place/list/content/index.js
@@ -1,5 +1,4 @@
 const { merge } = require('lodash');
-const { NACWO, NVS, SQP } = require('@asl/dictionary');
 const commonContent = require('../../content');
 
 module.exports = merge({}, commonContent, {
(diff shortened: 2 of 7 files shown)


pages/place/list/content/index.js
```diff
@@ -13,10 +14,10 @@ module.exports = merge({}, commonContent, {
   },
   fields: {
     nacwos: {
-      label: 'NACWOs'
+      label: `<abbr title="${NACWO}">NACWOs</abbr>`
```
lennym, line 17:
```suggestion
      label: `<Acronym>NACWO</Acronym>`
```

pages/place/list/schema/index.jsx
```diff
@@ -25,12 +28,14 @@ module.exports = {
...
-    filter: true
+    filter: true,
+    label: <span><Acronym>NACWO</Acronym>s</span>
   },
   nvssqps: {
     show: true,
     sortable: false,
-    filter: true
+    filter: true,
+    label: <span><Acronym>NVSs</Acronym>s / <Acronym>SQP</Acronym>s</span>
```
joefitter, line 38: this won't work - wrap it in a fragment

joefitter: ah sorry - just saw the span :D

lennym, line 38:
```suggestion
    label: <span><Acronym>NVS</Acronym>s / <Acronym>SQP</Acronym>s</span>
```

lennym: FWIW, unless the `<span>`s are needed for some reason I'd still use a `<Fragment>`

pages/place/formatters/index.jsx
```diff
@@ -1,15 +1,13 @@
...
 
 export default {
   suitability: {
     format: joinAcronyms,
-    label: defineValue,
     mapOptions: labelFromCode
   },
   holding: {
     format: joinAcronyms,
-    label: defineValue,
```
lennym, line 12: What is the reasoning behind removing these? It seems unrelated to the issue at hand.

wheelsandcogs: We changed the TableHeader component so that if the schema has a label prop present for a column, it tries to render it - in this case it would try and render the function without calling it, resulting in an error.

I'm not really sure what these were for - removing it doesn't change the header text (Suitability / Holding) because they fall back to using the content files.

lennym: Thanks.

PR #928, opened 2020-09-07
WIP Add contact us page

wheelsandcogs:
waiting on https://github.com/UKHomeOffice/asl-service/pull/201 bump


lennym:
If this page only appears in the external UI then it should be created as a page there. This is only for pages that appear to internal and external users.

There's no need to create an extra abstraction layer for things that don't need it.

wheelsandcogs:
Moving code back to asl

PR #930, opened 2020-09-08
Add page to provide a reason deadline has passed

wheelsandcogs:
~Needs the following to work:~
 * https://github.com/UKHomeOffice/asl-workflow/pull/309
 * https://github.com/UKHomeOffice/asl-public-api/pull/282


pages/task/read/routers/read.js
```diff
@@ -208,6 +208,16 @@ module.exports = () => {
     next();
   });
 
+  app.use((req, res, next) => {
+    if (req.task.data.model === 'project') {
+      const deadline = get(req.task, 'data.deadline');
+      const isExtended = get(deadline, 'isExtended', false);
+      const deadlineDate = get(deadline, isExtended ? 'extended' : 'standard');
+      res.locals.static.daysSinceDeadline = daysSinceDate(deadlineDate);
```
lennym, line 216: We already have `task.data.deadline` as an object for storing deadline-related metadata. Does it make sense for this to be a property on that object?

wheelsandcogs: Are you talking about adding it to the deadline object in the workflow decorator, or just patching it in the UI to avoid the extra locals var? All the current properties on the deadline object (standard, extended, isExtended, isExtendable) are static unless someone performs an action (e.g. resubmits the task / extends the deadline) - do we have any other `task.data.x` props that are dynamic?

pages/task/read/views/components/activity-log.jsx
```diff
@@ -27,15 +27,19 @@ function Action({ task, action, changedBy }) {
...
-  const deadline = get(task, 'data.deadline');
-  const deadlineDate = get(deadline, isExtended ? 'extended' : 'standard');
-  const daysSinceDeadline = daysSinceDate(deadlineDate, item.createdAt);
+  const deadlineAtTimeOfRecommendation = get(item, 'event.data.deadline');
+  let daysSinceDeadline;
+
+  if (deadlineAtTimeOfRecommendation) {
+    const isExtended = get(deadlineAtTimeOfRecommendation, 'isExtended', false);
+    const deadlineDate = get(deadlineAtTimeOfRecommendation, isExtended ? 'extended' : 'standard');
+    daysSinceDeadline = daysSinceDate(deadlineDate, item.createdAt);
```
wheelsandcogs, line 41: The "days since deadline" decoration does not get applied to the task data in activity items, so we still have to calc it here for the message to display.

PR #946, opened 2020-09-21
Restore correct headings to billing pages

lennym:
The top-level `title` property was already being used, so this needs to go back to how it was before to keep the correct page h1 heading.


lennym:
Note: this reverts part of https://github.com/UKHomeOffice/asl-pages/pull/945

PR #950, opened 2020-09-25
Feature - cat e pils

joefitter:
* Moved pil read to profile/id/pil - this contains the combined pil now
* Add a user to a trainig course
* Task page for cat e pils
* Updated pil procedures diff to include cat e


pages/pil/index.js
```diff
@@ -5,20 +5,22 @@ module.exports = settings => {
   const app = Router({ mergeParams: true });
 
   app.param('pilId', (req, res, next, pilId) => {
-    if (pilId === 'create') {
-      return next('route');
-    }
```
lennym, line 10: How come this is no longer needed?

pages/pil/index.js
```diff
@@ -5,20 +5,22 @@ module.exports = settings => {
...
+        req.pil.openTasks = response.json.meta.openTasks;
+        req.model = req.pil;
+        next();
       })
-      .then(() => next())
       .catch(next);
   });
 
+  app.use((req, res, next) => {
+    req.model = req.pil;
```
lennym, line 20: This is already done at L13 above, how come it needs to be assigned again.

pages/pil/index.js
lennym: I'm guessing because `/create is now mounted on `/pil` and not on `/pil/:id`.

pages/pil/unscoped/courses/index.js
```diff
@@ -13,7 +13,7 @@ module.exports = () => {
       .then(trainingCourse => {
         req.trainingCourseId = id;
         req.trainingCourse = trainingCourse;
-        req.model = trainingCourse;
+        req.model = res.locals.static.trainingCourse = trainingCourse;
```
lennym, line 16: Prefer obvious assignment.

```suggestion
        req.model = trainingCourse;
        res.locals.static.trainingCourse = trainingCourse;
```

pages/pil/index.js
joefitter: yup

pages/pil/index.js
joefitter: it should be here, not above, as req.pil is set in the profile/id router - I've removed it from L13

pages/profile/read/views/profile.jsx
```diff
@@ -33,12 +34,13 @@ class Profile extends React.Component {
 
     const profileRole = (establishments.find(est => est.id === estId) || {}).role;
     const pilIncomplete = pil && (pil.status === 'inactive' || pil.status === 'pending');
-    const pilActive = pil && pil.status === 'active';
+
+    const hasPil = pilLicenceNumber;
```
lennym, line 38: How does this work with the bug where we accidentally assigned licence numbers to admins applying for someone else?

There are a number (18 iirc) of profiles that have licence numbers but no related PILs.

pages/task/read/views/confirm.jsx
```diff
@@ -1,15 +1,19 @@
...
 
 const CommentForm = ({ task, values, errors, formFields }) => {
+  const model = task.data.model;
   let action = task.data.action;
   if (action === 'grant' && task.type === 'amendment') {
     action = 'update';
   }
   const title = <Snippet fallback={`status.${values.status}.action`}>{`status.${values.status}.action.${task.type}`}</Snippet>;
-  const requiresDeclaration = (task.data.model === 'pil' || task.data.model === 'project') && values.status === 'endorsed' && action !== 'review';
+  const requiresDeclaration = (model === 'pil' || model === 'project' || model === 'trainingPil') && values.status === 'endorsed' && action !== 'review';
```
lennym, line 14:
```suggestion
  const requiresDeclaration = ['pil', 'trainingPil', 'project'].includes(model) && values.status === 'endorsed' && action !== 'review';
```

pages/profile/read/views/profile.jsx
```diff
@@ -35,7 +35,7 @@ class Profile extends React.Component {
     const profileRole = (establishments.find(est => est.id === estId) || {}).role;
     const pilIncomplete = pil && (pil.status === 'inactive' || pil.status === 'pending');
 
-    const hasPil = pilLicenceNumber;
+    const hasPil = pilLicenceNumber && pil;
```
lennym, line 38: Is `pil` in this case a consolidated PIL record, and so there's no need here to take training PILs into account?

joefitter: yes

PR #962, opened 2020-09-30
Bugfix - licence number missing in task cat e view

joefitter:
* Updated schema of participant view to use correct accessor

### DIFF
--- a/pages/pil/unscoped/courses/participants/read/views/training-pil.jsx
+++ b/pages/pil/unscoped/courses/participants/read/views/training-pil.jsx
@@ -14,11 +14,22 @@ export default function TrainingPil({ className, trainingPil }) {
     'projectTitle',
     'projectId'
   ];
+
+  const modelSchema = {
+    ...pick({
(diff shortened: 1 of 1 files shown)


pages/pil/unscoped/courses/participants/read/views/training-pil.jsx
```diff
@@ -18,7 +18,7 @@ export default function TrainingPil({ className, trainingPil }) {
     <Inset className={className}>
       <ModelSummary
         model={{ ...trainingPil.trainingCourse, ...pick(trainingPil, 'expiryDate', 'issueDate') }}
-        schema={pick({ ...schema, ...trainingPilSchema }, fields)}
+        schema={{ ...pick({ ...schema, ...trainingPilSchema }, fields), projectId: { accessor: 'project.licenceNumber' } }}
```
lennym, line 21: This is quite a lot of code for an inline prop declaration. Could you move it out into a `const schema = ...`?

PR #966, opened 2020-10-01
Bugfix - use scoped est if set

joefitter:
* If req.establishment is set, use that for success page rather than task establishment, this shows the transfer to est for pil xfers

### DIFF
--- a/pages/success/index.js
+++ b/pages/success/index.js
@@ -120,7 +120,7 @@ module.exports = () => {
     merge(res.locals.static.content, { success });
     res.locals.static.taskId = req.taskId;
     res.locals.static.taskLabel = getTaskLabel(req.task);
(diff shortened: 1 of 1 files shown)


lennym:
What bug does this fix?

lennym:
PIL Transfer success pages show the wrong establishment?

joefitter:
yes

joefitter:
my words "this shows the transfer to est for pil xfers" were supposed to explain that, albeit, they weren't great words

lennym:
Hopefully this will also fix a slightly weird issue I spotted where the breadcrumb link text showed the receiving establishment, but the actual target of the link was the holding establishment. I think on the same page.

🤞

joefitter:
yeah it does fix that - I noticed that one too

PR #969, opened 2020-10-01
Only warn NTCOs about self endorsement

wheelsandcogs:
We were showing the message to non-NTCOs

### DIFF
--- a/pages/task/read/routers/read.js
+++ b/pages/task/read/routers/read.js
@@ -12,7 +12,7 @@ const updateData = require('../middleware/update-data');
 
 const endorsingOwnPil = (task, profile) => {
   const isNtco = !!profile.roles.find(r => r.type === 'ntco' && r.establishmentId === task.data.establishmentId);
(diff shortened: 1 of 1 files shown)


pages/task/read/routers/read.js
```diff
@@ -11,7 +11,8 @@ const { getEstablishment } = require('../../../common/helpers');
 const updateData = require('../middleware/update-data');
 
 const endorsingOwnPil = (task, profile) => {
-  return task.data.model === 'pil' && task.status === 'awaiting-endorsement' && profile.id === get(task, 'data.subject.id');
+  const isNtco = !!profile.roles.find(r => r.type === 'ntco' && r.establishmentId === task.data.establishmentId);
+  return task.data.model === 'pil' && task.status === 'awaiting-endorsement' && isNtco && profile.id === get(task, 'data.subject.id');
```
lennym, line 15:
```suggestion
  return isNtco && task.data.model === 'pil' && task.status === 'awaiting-endorsement' && profile.id === get(task, 'data.subject.id');
```

I don't know why, but having that condition right in the middle bothered me.

PR #971, opened 2020-10-02
Explain why the change ppl holder action is missing

wheelsandcogs:
Users are prevented from changing PPL holder while there are open tasks. Show a reason why instead of just hiding the link.

### DIFF
--- a/pages/project/read/views/index.jsx
+++ b/pages/project/read/views/index.jsx
@@ -368,7 +368,7 @@ export default function ProjectLandingPage() {
   const isEditable = model.status === 'active' || model.status === 'inactive';
   const grantedVersion = model.versions.find(v => v.status === 'granted');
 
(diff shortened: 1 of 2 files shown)


pages/project/read/views/index.jsx
```diff
@@ -368,7 +368,7 @@ export default function ProjectLandingPage() {
   const isEditable = model.status === 'active' || model.status === 'inactive';
   const grantedVersion = model.versions.find(v => v.status === 'granted');
 
-  const canChangeLicenceHolder = canUpdate && !openTask && isEditable && (!model.isLegacyStub || (model.isLegacyStub && asruLicensing));
+  const canChangeLicenceHolder = canUpdate && isEditable && (!model.isLegacyStub || (model.isLegacyStub && asruLicensing));
```
lennym, line 371: How come the `openTask` condition has been removed?

lennym: No problem, I think I see now. It's evaluated as a sub-condition if `canChangeLicenceHolder` is true.

PR #973, opened 2020-10-05
Feature - set isTrainingEst

joefitter:
* updated schema to include isTrainingEst
* updated task diff and confirm diff
* updated form router to format reveals as well as top level fields


lennym:
Setting this as WIP until we're ready to ship training PILs

PR #976, opened 2020-10-06
fix pil content for reapply

joefitter:


pages/profile/read/views/profile.jsx
```diff
@@ -44,6 +44,18 @@ class Profile extends React.Component {
...
 
+    let type;
+
+    if (pil && pil.status === 'revoked') {
+      type = 'pilReapply';
+    } else if (pilIncomplete) {
+      type = 'pilView';
+    } else {
+      type = 'pilApply';
+    }
```
wheelsandcogs, line 55: If you set `type = 'pilApply';` initially then you could omit this last condition but /shrug

PR #983, opened 2020-10-13
Remove redundant notice text about changes to areas

wheelsandcogs:


lennym:
What was the text, and can it also be removed from the content file?

wheelsandcogs:
Request was in a couple of tickets I brought in from the next sprint.

PR #987, opened 2020-10-19
keep tabs by default, add prop to configure

joefitter:


pages/task/list/views/tasklist.jsx
```diff
@@ -68,9 +70,24 @@ const Tasklist = ({
...
+
   return (
     <Fragment>
-      <StatusFilters tabs={tabs} progress={progress} />
+      {
+        displayTabs
+          ? (
+            <Fragment>
+              {
+                !!tabs.length && <Tabs active={selected}>
```
lennym, line 83: It might be good to add this condition to the component so we can get rid of the extra cruft to handle the (probably impossible) zero tab case..

PR #1014, opened 2020-11-11
Prevent removal of permissions from AA PPL holders

lennym:
If a person holds a PPL with an additional availability at an establishment then they should not be able to have permissions removed.


pages/profile/permission/index.js
```diff
@@ -22,8 +22,16 @@ module.exports = settings => {
...
+
+    const hasProjects = some(req.profile.projects, project => {
+      return project.status === 'active' && project.establishmentId === req.establishmentId;
+    });
+
+    const hasAdditionalProjects = some(req.profile.projects, project => {
+      return project.status === 'active' && project.additionalEstablishments.map(est => est.id).includes(req.establishmentId);
+    });
+
+    res.locals.static.isNamed = !!hasRoles || !!hasPil || !!hasProjects || !!hasAdditionalProjects;
```
wheelsandcogs, line 34: I'd prefer the cast to bool on assignment if we're using the "is/has" naming convention, but as it was already like this I'll ignore it.

PR #1035, opened 2020-11-26
Show PEL apply/amend declarations in activity log

wheelsandcogs:


pages/establishment/apply/content/index.js
```diff
@@ -42,7 +42,7 @@ module.exports = {
     }
   },
   fields: {
-    comments: {
+    comment: {
```
joefitter, line 45: what's the reason for changing this? Have you also made changes in workflow to ensure it is still added to the activity log?

wheelsandcogs: The comments were not being displayed in the activity log, this fixes that. I'm not sure if it worked at some point and then was broken, or if it's never displayed at all...

PR #1040, opened 2020-11-27
Allow insertion of newlines in establishment addresses

wheelsandcogs:


pages/establishment/formatters/index.jsx
```diff
@@ -9,6 +10,9 @@ const listFormatter = list => {
 };
 
 export default {
+  address: {
+    format: (value) => <span style={{ 'white-space': 'pre-line' }}>{value}</span>
```
joefitter, line 14: MATE

PR #1051, opened 2020-12-02
Changes to assigned named people do not need to be reviewed

wheelsandcogs:
Hide any diffs of named people assigned to places as they do not need to concern ASRU.


lennym:
Are changes to _only_ named people configured to resolve automatically?

wheelsandcogs:
> Are changes to _only_ named people configured to resolve automatically?

Yeah that already happens.

PR #1057, opened 2020-12-08
Feature - project csv

joefitter:
* added download csv link to projects list
* allow csv headers to be configuable
* added schema for project list csv

### DIFF
--- a/pages/project/schema/index.js
+++ b/pages/project/schema/index.js
@@ -1,28 +1,45 @@
 const { pick, merge } = require('lodash');
+const moment = require('moment');
+const { dateFormat } = require('../../../constants');
+
+function formatCSVDate(date) {
+  return date ? moment(date).format(dateFormat.long) : '-';
+}
 
 const schema = {
   id: {},
   title: {
-    show: true
+    show: true,
+    title: 'Licence title'
   },
   status: {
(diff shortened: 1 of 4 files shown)


pages/project/schema/index.js
```diff
@@ -35,12 +52,32 @@ const schema = {
...
+    toCSVString: formatCSVDate
   }
 };
 
-const getSchema = status => {
+const getSchema = (status, csv) => {
   const inactiveLicenceHolder = { licenceHolder: { label: 'Name' } };
 
+  if (csv) {
+    return merge({}, pick(schema, 'licenceNumber', 'title', 'licenceHolder', 'establishment', 'status', 'issueDate', 'amendedDate', 'expiryDate', 'species', 'raDate'));
```
lennym, line 78: What fields is this removing? I'm struggling to think why we would have a field in the online table and _not_ in the CSV?

joefitter: there's a list of fields on the ticket, createdAt and updatedAt are not on it (we use last amended date), also ID, probably others

lennym: 🤔  seems weird to have things in the digital view and NOT in the csv, but whatevs.

PR #1058, opened 2020-12-09
Use correct fields for experience when changing PPL holder

lennym:
Legacy PPLs use a different question for PPL holder experience, so the existing set of questions from new-style PPLs does not correctly update the PPL data.

Pass the PPL schema version to the schema generator function so it can return the correct set of experience questions depending on whether the PPL is legacy or new style.

### DIFF
--- a/pages/project/update-licence-holder/schema/experience-fields.js
+++ b/pages/project/update-licence-holder/schema/experience-fields.js
@@ -1,7 +1,23 @@
 const { flattenDeep, castArray } = require('lodash');
 const schema = require('@asl/projects/client/schema/v1/experience').default;
 
-module.exports = version => {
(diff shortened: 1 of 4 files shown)


pages/project/update-licence-holder/schema/experience-fields.js
```diff
@@ -1,7 +1,23 @@
...
+    return {
+      fields: [
+        {
+          name: 'experience-knowledge',
+          label: 'What relevant scientific knowledge or education do you have?',
+          alt: {
+            label: 'What relevant scientific knowledge or education does this person have?'
+          },
+          type: 'texteditor'
+        }
```
lennym, line 15: We _could_ get this from `@asl/projects` but for one field it feels like a giant faff.

PR #1060, opened 2020-12-09
Show projects with AA on profiles

wheelsandcogs:


lennym:
Does this handle old-style free text AA establishments?

wheelsandcogs:
> Does this handle old-style free text AA establishments?

It seems to work out of the box without any special code in the UI - must already be taken care of at API level.

Screenshot 2020-12-11 at 15 49 23

PR #1061, opened 2020-12-10
Feature - show last login to admins

joefitter:


pages/global-profile/views/index.jsx
```diff
@@ -134,8 +134,8 @@ export default function Index({ dedupe, AsruRolesComponent, children }) {
         {
           asruUser && (
             <Fragment>
-              <dt>Has login:</dt>
-              <dd>{ model.userId ? 'Yes' : 'No' }</dd>
+              <dt>Last activity:</dt>
+              <dd>{ model.lastLogin ? formatDate(model.lastLogin, dateFormat.datetime) : 'This user has not logged in' }</dd>
```
lennym, line 138: I think this need to be a _little_ more intelligent, so that we can still tell the people who have logged in, but not since we shipped the feature.

That is, if `lastLogin` is null, but `userId` is set then show `Unknown`, and only if both are null so `This user has not logged in`

joefitter: ok good point

PR #1066, opened 2020-12-14
Handle projects without `additionalEstablishments` set

lennym:


pages/profile/read/views/profile.jsx
```diff
@@ -75,23 +75,17 @@ class Profile extends React.Component {
...
-    });
+    }
 
-    const draftProjects = projects.filter(project => {
-      if (project.status !== 'inactive') {
-        return false;
-      }
-      const isPrimaryEstablishment = project.establishmentId === establishment.id;
-      const isAdditionalAvailability = project.additionalEstablishments.find(e => e.id === establishment.id);
-      return isPrimaryEstablishment || isAdditionalAvailability;
```
lennym, line 93: This logic was repeated twice with only the status changed, so consolidated into the `filterProjects` function above.

PR #1071, opened 2020-12-22
Add worker to do diff generation on server

lennym:
When retreiving previous values for a field from the server endpoint also attempt a diff calculation in a worker thread.

A sentence comparison is done first, then a word comparison.

If the worker thread does not resolve within 5 seconds then return whatever diff is available at that point. This means that for large answers that have been comprehensively re-written it is likely that only a sentence comparison will be available, but where only small changes exist in large answers a word-level comparison will be available.


lennym:
Needs https://github.com/UKHomeOffice/asl-projects/pull/665

PR #1073, opened 2021-01-06
The project manage tab should not display for inspectors

wheelsandcogs:


pages/project/read/views/index.jsx
```diff
@@ -66,6 +66,10 @@ export default function ProjectLandingPage() {
     delete sections.downloads;
   }
 
+  if (asruUser && asruInspector) {
+    delete sections.manage;
+  }
```
lennym, line 71: This feels like the wrong way round to approach this. For example in the case where a person has inspector and licensing roles it should remain visible.

Ensure that the tab _does_ appear for the people who _should_ see it, and don't include inspector on that list.

lennym:
If inspectors should not be able to perform PPL admin tasks then the permissions config should be reviewed as well.

https://github.com/UKHomeOffice/asl-permissions/blob/master/config.js#L83

I would potentially recommend using a consolidated permissions call to determine visibility of the manage tab. That is, if a user has at least one of a list of permissions on the project then the tab is visible, otherwise it is hidden. That would avoid spreading the logic into various places.

lennym:
I think specifically if all [these permissions checks](https://github.com/UKHomeOffice/asl-pages/blob/master/pages/project/read/index.js#L42) come in false then the Manage tab should be hidden.

PR #1081, opened 2021-01-20
Add a link to download the csv of all approved areas at est

wheelsandcogs:
~waiting on https://github.com/UKHomeOffice/asl-components/pull/187~

### DIFF
--- a/pages/place/list/views/index.jsx
+++ b/pages/place/list/views/index.jsx
@@ -59,10 +59,6 @@ export default function Places() {
   schema.nacwos.label = <Fragment><Acronym>NACWO</Acronym>s</Fragment>;
   schema.nvssqps.label = <Fragment><Acronym>NVS</Acronym>s / <Acronym>SQP</Acronym>s</Fragment>;
 
-  const otherLinks = [
-    <Link key="csv-download" page="place.list" query={{csv: 1}} label="Download as CSV" />
-  ];
-
   return (
     <Fragment>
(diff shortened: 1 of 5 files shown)


pages/place/list/views/index.jsx
```diff
@@ -59,6 +59,14 @@ export default function Places() {
   schema.nacwos.label = <Fragment><Acronym>NACWO</Acronym>s</Fragment>;
   schema.nvssqps.label = <Fragment><Acronym>NVS</Acronym>s / <Acronym>SQP</Acronym>s</Fragment>;
 
+  const actions = <Fragment>
+    <Link page="place.list" query={{csv: 1}} label={<Snippet>actions.download</Snippet>} />
```
lennym, line 63: One thing to note here is that it will remove any filters that have been applied and always return the full set.

I'm not sure if this is intentional or not.

wheelsandcogs: I believe that's what was intended - isn't that the case with all of our CSV downloads (i.e. return the full set)? It may not be obvious from the UI what to expect however.

wheelsandcogs: Confirmed with Ben - link text now says "Download all as CSV" rather than just "Download as CSV"

lennym: For the record, the other places we have CSV downloads (project content search, perf metrics) respect whatever filters have been applied.

PR #1087, opened 2021-01-25
Bugfix - how to even switch

joefitter:
* projects used to be cause by default, not anymore since a projects case was added


pages/success/index.js
```diff
@@ -36,7 +36,7 @@ const getTaskLabel = task => {
       if (action === 'grant-ra') {
         return 'Retrospective assessment';
       }
-      break;
+      return `${upperFirst(model)} ${taskType}`;
```
lennym, line 39:
```suggestion
```

PR #1090, opened 2021-01-25
dont show start ra banner on active project

joefitter:


lennym:
We probably also want to add a check on the POST to start the RA to throw a 400

PR #1098, opened 2021-02-02
Update wording for PIL E revoke UI

wheelsandcogs:


lennym:
If there is not already an integration test for this could you add one?

PR #1109, opened 2021-02-22
Add version holder to state passed to projects

wheelsandcogs:
Licence holder for a version now needs to come from the version (not the version data) instead of the project, so we need to add it to the state used by asl-projects.

Needs:
 * ~https://github.com/UKHomeOffice/asl-projects/pull/693~
 * ~https://github.com/UKHomeOffice/asl-public-api/pull/312~


pages/project-version/update/js/index.js
```diff
@@ -34,12 +34,13 @@ start({
...
     establishments: state.static.user.asruUser
-      ? state.static.project.licenceHolder.establishments
+      ? state.model.licenceHolder.establishments
       : state.static.user.establishments.map(e => ({ id: e.id, name: e.name })),
     showConditions: state.static.showConditions,
     editConditions: state.static.editConditions,
     raCompulsory: state.model.raCompulsory,
-    training: state.static.training
+    training: state.static.training,
+    licenceHolder: state.static.licenceHolder
```
lennym, line 43: It appears from this line combined with the change at L37 that the licence holder is at `state.static.licenceHolder` _and_ `state.model.licenceHolder`.

Is there a meaningful difference between those two things?

lennym: If there isn't then I'd have a strong preference to remove the one from `static` and one serve the value once to avoid unnecessary data being send in the page payload.

pages/retrospective-assessment/update/js/index.js
```diff
@@ -30,7 +30,8 @@ start({
     grantedVersion: state.static.grantedVersion,
     newApplication: state.static.newApplication,
     sidebarLinks: state.static.sidebarLinks,
-    schemaVersion: 'RA'
+    schemaVersion: 'RA',
+    licenceHolder: state.static.project.licenceHolder
```
lennym, line 34: Does this exist?

wheelsandcogs: Yes, `project.licenceHolder` is always the current licence holder for the project, I have not removed this. I believe the RA only applies to the most recent granted version, so this is correct?

What I removed from the API was `version.project.licenceHolder' because when looking at specific versions we should use `version.licenceHolder` instead.

pages/project-version/update/js/index.js
wheelsandcogs: I'm looking at this atm

PR #1111, opened 2021-02-26
Remove invisible unicode from editor values in comparison

lennym:
Invisible unicode characters like `\uFEFF` cause word boundaries to be added where none actually exist, and string lengths to be miscalculated.

Remove any invisible unicode control characters from text before performing comparison to ensure that the changes render correctly.

~WIP pending projects bump~

### DIFF
--- a/pages/project-version/middleware/diff/compare.js
+++ b/pages/project-version/middleware/diff/compare.js
@@ -1,0 +1,80 @@
+const { Value } = require('slate');
+const { diffWords, diffSentences, diffArrays } = require('diff');
+const last = require('lodash/last');
+
+const parseValue = (val) => {
+  if (typeof val === 'string') {
+    val = JSON.parse(val || '{}');
+  }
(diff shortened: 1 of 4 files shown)


pages/project-version/middleware/diff/compare.js
```diff
@@ -0,0 +1,82 @@
...
+        before = parseValue(a);
+        after = parseValue(b);
+      } catch (e) {
+        return { error: e, added: [], removed: [] };
+      }
+
+      if (granularity === 'word') {
+        diffs = diffWords(normaliseWhitespace(before.document.text), normaliseWhitespace(after.document.text));
+      } else {
+        diffs = diffSentences(normaliseWhitespace(before.document.text), normaliseWhitespace(after.document.text));
```
lennym, line 48: The addition of `normaliseWhitespace` in the two lines above is the significant change here.

PR #1113, opened 2021-03-05
pass the RA reasons to the client

wheelsandcogs:


lennym:
I notice this is for the update script. Is there a requirement to also play back reasons on the readonly view?

wheelsandcogs:
Good question... I'll submit one and find out.

PR #1116, opened 2021-03-15
Feature - ROPS

joefitter:
* Bumped react components and asl components with updates to species selector, and single radio options
* Added ROPs journey - create, update, submit
* Added procs journey - create, update, delete
* updated implementaions of existing species selector


pages/rops/procedures/list/views/index.jsx
```diff
@@ -0,0 +1,153 @@
...
+        <button className="link" onClick={onClick}>Delete</button>
+      </form>
+    </Fragment>
+  );
+}
+
+export default function Procedures() {
+  const { project, rop } = useSelector(state => state.static);
+  const editable = rop.status === 'draft';
+  console.log({ editable });
```
lennym, line 111: OHAI!

pages/rops/procedures/list/views/index.jsx
```diff
@@ -0,0 +1,153 @@
...
+              page="rops.procedures.create"
+              label={<Snippet>procedures.add</Snippet>}
+            />
+          </Fragment>
+        )
+      }
+      {
+        rop.procedures.length
+          ? (
+            <div style={{ overflowX: 'scroll', maxWidth: '2000px' }}>
```
lennym, line 144: What's the source of this magic number?

pages/rops/procedures/list/views/index.jsx
lennym:
```suggestion
```

pages/rops/procedures/list/views/index.jsx
lennym: Am I right in thinking that this caps the total width of the table at 2000px even if the user is viewing on a super-wide monitor?

Are we sure that's a good thing? Letting it go to viewport width however large that might be seems potentially better

joefitter: not actually sure, let me try without

PR #1121, opened 2021-03-18
Remove rops tab for non licensing. Bump things

joefitter:


lennym:
I'm ok with this _for now_ but I think eventually we'll probably want to attach this to a `project.update` permission.

pages/project/read/views/index.jsx
```diff
@@ -74,7 +74,7 @@ export default function ProjectLandingPage() {
     delete sections.history;
   }
 
-  if (project.status === 'inactive') {
+  if (project.status === 'inactive' || !asruLicensing) {
```
lennym, line 77: I think this needs a few more conditions to cover AA and the like also. I think on L71 also get `canUpdate` and add a `|| !canUpdate` here too.

lennym: I know that will be a no-op _right now_, but it means when we do the story to not lock it down to licensing only then we won't need to add anything new so there'll be no risk of exposing it too widely.

pages/project/read/views/index.jsx
```diff
@@ -68,13 +68,13 @@ export default function ProjectLandingPage() {
...
 
-  const { additionalAvailability, showManageSection } = useSelector(state => state.static);
+  const { additionalAvailability, showManageSection, canUpdate } = useSelector(state => state.static);
 
   if (additionalAvailability || (!hasHistory() && !hasPreviousVersions())) {
     delete sections.history;
   }
 
-  if (project.status === 'inactive') {
+  if (project.status === 'inactive' && canUpdate) {
```
lennym, line 77:
```suggestion
  if (project.status === 'inactive' || !canUpdate || !asruLicensing) {
```

lennym: This will need the `asruLisensing` bit putting back to wherever that was before.

pages/project/read/views/index.jsx
```diff
@@ -52,7 +52,7 @@ function hasPreviousVersions() {
 
 export default function ProjectLandingPage() {
   const project = useSelector(state => state.model);
-  const { openRaTask, url } = useSelector(state => state.static);
+  const { openRaTask, url, isLicensing } = useSelector(state => state.static);
```
lennym, line 55:
```suggestion
  const { openRaTask, url } = useSelector(state => state.static);
```

pages/project/read/views/index.jsx
```diff
@@ -68,13 +68,13 @@ export default function ProjectLandingPage() {
     downloads: <Snippet>{`${snippetPath}.downloads`}</Snippet>
   };
 
-  const { additionalAvailability, showManageSection } = useSelector(state => state.static);
+  const { additionalAvailability, showManageSection, canUpdate } = useSelector(state => state.static);
```
lennym, line 71:
```suggestion
  const { additionalAvailability, showManageSection, canUpdate, isLicensing } = useSelector(state => state.static);
```

PR #1134, opened 2021-03-26
View all rops due at an establishment for a specific year

wheelsandcogs:
~needs https://github.com/UKHomeOffice/asl-public-api/pull/318~

### DIFF
--- a/pages/common/content/index.js
+++ b/pages/common/content/index.js
@@ -11,7 +11,7 @@ module.exports = {
       update: 'Amend establishment details',
       apply: 'Apply for licence',
       fees: 'Estimated licence fees',
-      rops: 'Return of procedures'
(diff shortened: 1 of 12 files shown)


pages/common/content/index.js
```diff
@@ -10,7 +10,8 @@ module.exports = {
       read: 'Details',
       update: 'Amend establishment details',
       apply: 'Apply for licence',
-      fees: 'Estimated licence fees'
+      fees: 'Estimated licence fees',
+      rops: 'Return of procedures'
```
lennym, line 14: It's `Returns` (plural) below. Should it also be here?

wheelsandcogs: yes

PR #1138, opened 2021-03-30
fix typo

wheelsandcogs:


pages/rops/update/content/index.js
```diff
@@ -105,7 +105,7 @@ Octopus and squid should be counted as free feeding from when they hatch, and cu
 Select all that apply`,
       options: {
         'uk-licenced': {
-          label: 'In the UK at licenced establishment',
+          label: 'In the UK at a licenced establishment',
```
lennym, line 108: I think this should also be `licensed`

wheelsandcogs: I was wondering that but the ticket said licenced

PR #1140, opened 2021-03-31
Filter to active aa when licence is granted

wheelsandcogs:


lennym:
Can we get an integration test for this one too?

PR #1141, opened 2021-03-31
Feature - unsubmit rop

joefitter:
* added unsubmit rop option if rop is submitted

### DIFF
--- a/pages/rops/procedures/index.js
+++ b/pages/rops/procedures/index.js
@@ -5,7 +5,7 @@ module.exports = () => {
   const app = Router({ mergeParams: true });
 
   app.param('procedureId', (req, res, next, procedureId) => {
-    if (procedureId === 'create') {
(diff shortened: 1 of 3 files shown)


pages/rops/procedures/index.js
```diff
@@ -17,6 +17,17 @@ module.exports = () => {
       .catch(next);
   });
 
+  app.post('/unsubmit', (req, res, next) => {
+    const params = {
+      method: 'POST'
+    };
+    req.api(`/establishment/${req.establishmentId}/project/${req.projectId}/rop/${req.ropId}/unsubmit`, params)
+      .then(() => {
+        res.redirect(req.buildRoute('rops.procedures'));
```
lennym, line 26: Do we want to add a toaster message here?

lennym: a) Yes
b) `Return updated.`

joefitter: 👍

PR #1144, opened 2021-04-08
WIP Ignore empty array props added to protocols when checking for changes

wheelsandcogs:


lennym:
I feel like there's some solid benefit in adding a bit of unit test coverage around this area. We're starting to do pretty complex data interrogation and just kind of assuming it works fine.

Would perhaps be good to add some basic coverage of `getChanges`

wheelsandcogs:
Instead of fixing the symptoms (ignoring certain changes to protocols) fix the actual root cause (don't add protocol sections that legacy licences don't use): https://github.com/UKHomeOffice/asl-projects/pull/715

PR #1145, opened 2021-04-12
Display reasons for RA in NTS when required

wheelsandcogs:
~waiting on https://github.com/UKHomeOffice/asl-projects/pull/716~

### DIFF
--- a/pages/project-version/nts/views/components/retrospective-decision.jsx
+++ b/pages/project-version/nts/views/components/retrospective-decision.jsx
@@ -1,15 +1,26 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import RAContent from '@asl/projects/client/constants/retrospective-assessment';
+import RaReasons from '@asl/projects/client/components/ra-reasons';
 
-export default function RetrospectiveDecision({ version }) {
+export default function RetrospectiveDecision({ project, version }) {
   const raCompulsory = version.raCompulsory;
   const raRequired = !!version.retrospectiveAssessment;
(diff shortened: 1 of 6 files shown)


pages/project-version/nts/views/components/retrospective-decision.jsx
```diff
@@ -1,15 +1,26 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import RAContent from '@asl/projects/client/constants/retrospective-assessment';
+import RaReasons from '@asl/projects/client/components/ra-reasons';
 
-export default function RetrospectiveDecision({ version }) {
+export default function RetrospectiveDecision({ project, version }) {
   const raCompulsory = version.raCompulsory;
   const raRequired = !!version.retrospectiveAssessment;
+  const isRequired = raCompulsory || raRequired || project.raDate;
```
lennym, line 8: What's the logic behind adding the 3rd clause?

Why would the first two not be sufficient?

wheelsandcogs: This was to make sure the message on the current NTS still says there is an RA required if the current version does not have anything that triggers `raCompulsory` or `raRequired`.

lennym: I had in my head that those are now data flags and not calculated (and also that once `true` they remain true even if amended) so this _shouldn't_ be an issue.

wheelsandcogs: It's possible that I only needed it for seeded test projects then, but no harm leaving it in.

PR #1150, opened 2021-04-15
fix req.project was not set so hasNhps() could not see project species

wheelsandcogs:


pages/rops/index.js
```diff
@@ -8,6 +8,7 @@ module.exports = () => {
     req.api(`/establishment/${req.establishmentId}/project/${req.projectId}/rops/${req.ropId}`)
       .then(({ json: { data, meta } }) => {
         req.rop = data;
+        req.project = req.rop.project;
```
wheelsandcogs, line 11: this is the actual bugfix.

PR #1154, opened 2021-04-16
Rename licence labels

wheelsandcogs:


pages/common/routers/related-tasks.js
```diff
@@ -49,7 +49,7 @@ module.exports = getQuery => {
       next();
     },
     locals: (req, res, next) => {
-      res.locals.static.content = merge({}, res.locals.static.content, content);
+      res.locals.static.content = merge({}, content, res.locals.static.content);
```
wheelsandcogs, line 52: Need to change the merge order here, otherwise the content from relatedTasks overrides `fields.establishment.label` and we can't change it in `project/read/content/index.js`

pages/project-version/downloads/content/index.js
```diff
@@ -2,10 +2,10 @@ module.exports = {
   title: 'Downloads',
   licence: {
     inactive: {
-      heading: 'Draft licence',
-      link: 'Draft licence: {{title}} (.pdf)',
+      heading: 'Licence preview',
+      link: 'Licence preview: {{title}} (.pdf)',
       hint: 'Download a preview of how the licence will appear if granted. This doesn\'t include any restrictions the Home Office may add.',
-      back: 'Back to draft licence'
+      back: 'Back to licence preview'
```
lennym, line 8: Is this particular change specced? It seems like the thing they'd be going back to at this point still _is_ the draft licence since it refers to the editing UI and not the PDF download.

pages/common/routers/related-tasks.js
lennym: Makes sense. This way round sets defaults for the related tasks, but prefers page-local content definitions.

pages/project-version/downloads/content/index.js
wheelsandcogs: It is not, thought it'd just been missed as we were changing draft licence -> licence preview. I'll revert it.

PR #1176, opened 2021-05-04
CSV should contain projects of any status

wheelsandcogs:


pages/project/list/index.js
```diff
@@ -36,9 +36,8 @@ module.exports = settings => {
       next();
     },
     getApiPath: (req, res, next) => {
-      const query = {
-        status: req.query.status || 'active'
-      };
+      const allStatuses = ['inactive', 'pending', 'active', 'expired', 'revoked', 'transferred'];
```
wheelsandcogs, line 39: I looked at just not passing the status when it's a CSV download, but we default to 'active' in multiple places all the way up the stack, and I don't want to break anything that's assuming those defaults are in place.

PR #1217, opened 2021-05-26
Feature - collab perms

joefitter:
* Moved collabs to own tab on project read page
* Removed delete router and action - replaced with edit interface
* Updated add interface to include role field


joefitter:
WIP until new perms tested

PR #1242, opened 2021-06-09
Add default value to species validator

lennym:
Without a default value then it is not possible to return to a legacy project ROP if it was abandoned at the species selector step because it throws an error trying to call `Object.values(undefined)`.


joefitter:
this is a very good programming 💯 🥇

PR #1248, opened 2021-06-14
Add new UI for removing permissions from a user

lennym:
Adds an intermediate confirmation page with a notice about deleting draft applications.

~Note: No further work to be done here, but WIP pending the code in resolver to actually do the deletion of draft applications.~

### DIFF
--- a/pages/profile/remove/views/index.jsx
+++ b/pages/profile/remove/views/index.jsx
@@ -1,0 +1,46 @@
+import React, { Fragment } from 'react';
+import { useSelector, shallowEqual } from 'react-redux';
+import { Header, Link, Snippet, Form } from '@asl/components';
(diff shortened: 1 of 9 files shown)


pages/profile/remove/views/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+                <ul>
+                  {
+                    drafts.map(project => {
+                      return <li key={project.id}>
+                        <Link page="project.read" projectId={project.id} label={project.title} />
+                      </li>;
+                    })
+                  }
+                </ul>
+                <p>Applications that need to be kept should be downloaded or transferred to another person first.</p>
```
wheelsandcogs, line 36: Any reason there's a few bits of content not in snippets?

PR #1249, opened 2021-06-14
Fix bug with granted version lookup

lennym:
* Handle case where the previous granted version does not exist, and error was thrown reading id property. Instead resolve early.
* Handle lookup of granted RA correctly

### DIFF
--- a/pages/project-version/middleware/index.js
+++ b/pages/project-version/middleware/index.js
@@ -183,14 +183,20 @@ const getPreviousVersion = (req, type = 'project-versions') => {
 };
 
 const getGrantedVersion = (req, type = 'project-versions') => {
-  const key = type === 'project-versions' ? 'granted' : 'grantedRa';
-  if (!req.project || !req.project[key]) {
+  if (!req.project) {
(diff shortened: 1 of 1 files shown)


pages/project-version/middleware/index.js
```diff
@@ -183,14 +183,20 @@ const getPreviousVersion = (req, type = 'project-versions') => {
...
 const getGrantedVersion = (req, type = 'project-versions') => {
-  const key = type === 'project-versions' ? 'granted' : 'grantedRa';
-  if (!req.project || !req.project[key]) {
+  if (!req.project) {
+    return Promise.resolve();
+  }
+  const key = type === 'project-versions' ? 'versions' : 'retrospectiveAssessments';
+  const model = type === 'project-versions' ? 'version' : 'retrospectiveAssessment';
+
+  const granted = req.project[key]
```
wheelsandcogs, line 192: Can we be sure `req.project[key]` always exists here?

lennym: Probably, but I've added a check just in case.

PR #1253, opened 2021-06-18
Render disabled button if user cannot be removed from an establishment

lennym:
Adding a disabled prop to the link doesn't change its behaviour - if the user clicks on the disabled link they are still taken to the remove page.

Instead of the link, render a disabled button when the functionality is required to be switched off.


wheelsandcogs:
it-works

PR #1262, opened 2021-06-25
Optimise data payloads for ROPs

lennym:
The complete version data is being included in three different places in `INITIAL_STATE` which causes very slow load times for large projects.

None of these need to be sent if we move the basic calculations being done on the client to the server and send calculated props as required.

Also remove duplication of `state.model` and `state.static.rop` to further reduce redundant data in the payload.

### DIFF
--- a/pages/rops/update/views/index.jsx
+++ b/pages/rops/update/views/index.jsx
@@ -6,7 +6,7 @@ import formatters from '../formatters';
 import components from './components';
 
 export default function Step() {
-  const { step, project, schema } = useSelector(state => state.static);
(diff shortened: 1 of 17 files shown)


pages/rops/update/views/index.jsx
```diff
@@ -6,7 +6,7 @@ import formatters from '../formatters';
 import components from './components';
 
 export default function Step() {
-  const { step, project, schema } = useSelector(state => state.static);
+  const { step, schema, project } = useSelector(state => state.static);
```
lennym, line 9: COOL CHANGE BRO!

PR #1264, opened 2021-06-25
Show extra context in reveals for severity options

wheelsandcogs:
~waiting on https://github.com/UKHomeOffice/asl-components/pull/213~

### DIFF
--- a/pages/common/routers/form.js
+++ b/pages/common/routers/form.js
@@ -101,6 +101,36 @@ const getConditionalRevealKeys = schema => chain(schema)
   .flatten()
   .value();
 
+const flattenDetailsReveals = schema => {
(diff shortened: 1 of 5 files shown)


pages/common/routers/form.js
```diff
@@ -204,6 +234,9 @@ module.exports = ({
   };
 
   const _process = (req, res, next) => {
+    // if we don't clone the schema here it gets modified in place by flattenDetailsReveals() because object refs...
+    // thought it would be clearer with an explicit assignment
+    req.form.schema = flattenDetailsReveals(cloneDeep(req.form.schema));
```
lennym, line 239: It's probably safer to do this inside the function

wheelsandcogs: its recursive tho...

PR #1270, opened 2021-06-29
Feature - rops back link

joefitter:
* Added back link to rops pages
* bumped components with custom cancel link

### DIFF
--- a/pages/rops/components/cancel-link.jsx
+++ b/pages/rops/components/cancel-link.jsx
@@ -1,0 +1,16 @@
+import React from 'react';
+
+export default function CancelLink() {
+  if (typeof window !== 'undefined') {
+    if (window.history.length <= 1) {
+      return null;
+    }
+  }
+
+  function onClick(e) {
+    e.preventDefault();
+    window.history.back();
+  }
+
+  return <a href="#" onClick={onClick}>Back</a>;
+}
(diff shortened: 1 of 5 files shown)


pages/rops/components/cancel-link.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+export default function CancelLink() {
+  if (typeof window !== 'undefined') {
+    if (window.history.length <= 1) {
+      return null;
+    }
+  }
```
lennym, line 8: We probably want to change this section to `if (!window || window.history.length <= 1) {` so that it also returns null when server rendering.

joefitter: if (!window) would throw an error on server as undefined, you need to use typeof window !== undefined (see line above)

lennym: In which case `if (typeof window === 'undefined' || window.history.length <= 1) {`

The key being that we don't fall through to render a potentially broken `<a>` when rendering on the server.

joefitter: yeah currently its rendered on the server then killed by the client. I'll update it to return null from server

lennym: Fanks.

lennym (approved):

PR #1285, opened 2021-07-08
Bugfix - rop nil return logic

joefitter:
If a nil-return is triggered then journey abandonned, the logic to determine which step the user should be redirected to doesn't take target into account.

* Updated logic for nil return to redirect if either condition is true
* Updated multi-step router logic to apply target function of previous step if present


lennym:
Can you add an integration test for this case?

joefitter:
👍

PR #1295, opened 2021-07-27
force strings to display as strings in CSVs

wheelsandcogs:
Excel and other spreadsheet apps commonly used to open CSVs, try to "helpfully" format strings that vaguely look like numbers as actual numbers which can lead to weird outputs.

### DIFF
--- a/pages/common/routers/datatable.js
+++ b/pages/common/routers/datatable.js
@@ -158,7 +158,7 @@ module.exports = ({
           if (value && /[0-9]+/.test(value) && !value.includes('"')) {
             return { value: `="${value}"`, quote: false };
           } else {
-            return { value, quote: false };
+            return value;
           }
         }
       },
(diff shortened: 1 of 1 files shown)


pages/common/routers/datatable.js
```diff
@@ -153,6 +153,15 @@ module.exports = ({
     const stringifier = csv({
       bom: true,
       header: true,
+      cast: {
+        string: (value) => {
+          if (value && /[0-9]+/.test(value) && !value.includes('"')) {
+            return { value: `="${value}"`, quote: false };
+          } else {
+            return { value, quote: false };
```
lennym, line 161: What is `quote: false` doing here?

lennym: Reading the docs, just reverting to the default seems safest.
```suggestion
            return value;
```

PR #1299, opened 2021-07-30
Fix: legacy licences were still displaying the "AMENDED" badge after an amendment was granted

wheelsandcogs:


pages/project-version/read/js/index.js
```diff
@@ -13,7 +13,7 @@ start({
     id: state.model.id
   }),
   comments: state.static.comments,
-  changes: state.static.isGranted
+  changes: state.static.isGranted || state.static.legacyGranted
```
lennym, line 16:
```suggestion
  changes: (state.static.isGranted || state.static.legacyGranted)
```

PR #1300, opened 2021-08-02
if other species are present, don't preselect any species

wheelsandcogs:
If a project has "other" species, then they must manually select all of the species in the ROPs, we don't pre-select any for them.

### DIFF
--- a/pages/rops/update/schema/index.js
+++ b/pages/rops/update/schema/index.js
@@ -37,7 +37,8 @@ module.exports = req => {
         inputType: 'speciesSelector',
         projectSpecies: true,
         presets: (req.version.data.species || []).find(s => s.includes('other'))
-          ? [] // if any "other" type species, don't preselect anything
+          // don't add species from project if "other" values are selected
+          ? req.rop.procedures.map(p => p.species)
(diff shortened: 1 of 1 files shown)


pages/rops/update/schema/index.js
```diff
@@ -36,10 +36,12 @@ module.exports = req => {
       species: {
         inputType: 'speciesSelector',
         projectSpecies: true,
-        presets: [
-          ...(req.version.data.species || []).filter(s => !s.includes('other')),
-          ...(req.rop.procedures.map(p => p.species))
-        ],
+        presets: (req.version.data.species || []).find(s => s.includes('other'))
+          ? [] // if any "other" type species, don't preselect anything
```
lennym, line 40: I think we still need to preselect the values from `req.rop.procedures.map(p => p.species)` so that soecies which have been previously added to species cannot be removed.
```suggestion
          // don't add species from project if "other" values are selected
          ? req.rop.procedures.map(p => p.species)
```

wheelsandcogs: Thanks... missed that.

PR #1309, opened 2021-08-12
Bugfix - dont include nhps qs if not in rop

joefitter:
* Abstracted getSpecies function
* Use this for list of species to see if nhps are included


wheelsandcogs:
Fancy swapping out this one for the helper as well? 

https://github.com/UKHomeOffice/asl-pages/blob/master/pages/rops/update/views/components/confirm.jsx#L68-L70

PR #1312, opened 2021-08-16
Feature - redirect to confirm

joefitter:
* Updated auto-continue logic to only redirect on submit of confirm page
* Updted button text if there is a redirect step to be completed

### DIFF
--- a/pages/common/routers/multi-step.js
+++ b/pages/common/routers/multi-step.js
@@ -41,17 +41,14 @@ module.exports = ({ schema, config, root, postData = (req, res, next) => next() 
   });
 
   app.use((req, res, next) => {
-    const step = req.step;
     const allSteps = Object.keys(config);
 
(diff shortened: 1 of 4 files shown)


pages/common/routers/multi-step.js
```diff
@@ -41,17 +41,14 @@ module.exports = ({ schema, config, root, postData = (req, res, next) => next()
...
   app.use((req, res, next) => {
-    const step = req.step;
     const allSteps = Object.keys(config);
 
-    const previousSteps = allSteps.slice(0, allSteps.indexOf(step));
-
-    const includedSteps = previousSteps.filter(s => {
+    const steps = allSteps.filter(s => {
       const include = config[s].include;
       return !include || include(req);
```
lennym, line 48: This implies that setting `include: false` will cause a step be be included. Seems odd.

lennym: Sorry for being "commenting on white lines in the PR guy"

PR #1317, opened 2021-08-18
Bugfix - nil return

joefitter:
* Updated confim page submit logic to redirect to nil return if applic
* fixed postnatal confirm page label

### DIFF
--- a/pages/rops/update/views/components/confirm.jsx
+++ b/pages/rops/update/views/components/confirm.jsx
@@ -107,7 +107,9 @@ export default function Confirm() {
           <dd>{yn(rop.proceduresCompleted)}</dd>
 
           <dt>Only protected embryonic forms used</dt>
-          // need to reverse the boolean for postnatal animals because the phrasing of the question changes
+          {
+            // need to reverse the boolean for postnatal animals because the phrasing of the question changes
+          }
           <dd>{yn(rop.postnatal === null ? null : !rop.postnatal)}</dd>
 
           <dt>Endangered animals used</dt>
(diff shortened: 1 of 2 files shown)


pages/rops/update/views/components/confirm.jsx
```diff
@@ -107,7 +107,7 @@ export default function Confirm() {
           <dd>{yn(rop.proceduresCompleted)}</dd>
 
           <dt>Only protected embryonic forms used</dt>
-          <dd>{yn(rop.postnatal)}</dd>
+          <dd>{yn(rop.postnatal === null ? null : !rop.postnatal)}</dd>
```
lennym, line 110: How come this particular field gets this treatment and not all the other `yn` fields that might be null?

lennym:
To clarify, if someone goes back to an incomplete return which would be a nil return they are taken to a mostly empty confirm page, and then on hitting the continue button they go to the nil return screen?

joefitter:
yup

pages/rops/update/views/components/confirm.jsx
joefitter: `yn` function takes null into account, however the boolean is switched for this field (for reasons) so `true` needs to show `'Yes'`, `false` `'No'` and `null` `'-'`

joefitter: we could do something like `typeof rop.postnatal === 'boolean' ? !rop.postnatal : rop.postnatal` if you think it's clearer

lennym:
```suggestion
          // need to reverse the boolean for postnatal animals because the phrasing of the question changes
          <dd>{yn(rop.postnatal === null ? null : !rop.postnatal)}</dd>
```

PR #1319, opened 2021-08-19
Legacy PPL / PPL transfer AWERB fixes

wheelsandcogs:
This PR fixes two issues:
1. Legacy PPL amendments submitted by non-PELH/HOLCs were not correctly capturing the legacy AWERB text. This should now be captured correctly regardless of who submits the amendment.

2. PPL transfers were not asking the receiving establishment for their AWERB review date. We now force the receiving establishment to enter their AWERB review date (they cannot be exempt) and also add / review any Additional Availability AWERB dates

### DIFF
--- a/pages/project-version/update/submit/schema/index.js
+++ b/pages/project-version/update/submit/schema/index.js
@@ -13,7 +13,7 @@ const getDateField = establishmentName => {
   };
 };
 
-const getAwerbQuestion = ({ isLegacy, isAmendment, awerbEstablishments }) => {
+const getAwerbQuestion = ({ isLegacy, canBeAwerbExempt, awerbEstablishments }) => {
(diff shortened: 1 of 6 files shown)


pages/project-version/update/submit/schema/index.js
```diff
@@ -13,7 +13,7 @@ const getDateField = establishmentName => {
   };
 };
 
-const getAwerbQuestion = ({ isLegacy, isAmendment, awerbEstablishments }) => {
+const getAwerbQuestion = ({ isLegacy, canBeAwerbExempt, awerbEstablishments }) => {
```
wheelsandcogs, line 16: `isAmendment` previously included transfers in certain circumstances, but transfers cannot be exempt.

PR #1323, opened 2021-08-23
added confirm edit page

joefitter:


pages/rops/update/confirm/views/index.jsx
```diff
@@ -0,0 +1,26 @@
...
+      />
+      <Snippet>content</Snippet>
+      <div className="control-panel">
+        <Link
+          page="rops.update"
+          label="Continue"
+          className="govuk-button"
+          step={step}
+        />
+        <CancelLink />
```
lennym, line 22: Where does this link to? And does it care whether you edited from the summary page or the procedures list?

joefitter: it does browser.back() - we use it on all rops pages

PR #1343, opened 2021-09-02
Improved tasklist filters

wheelsandcogs:
Adds a new "PPL type" filter that becomes available when filtering to just PPL tasks.

before

after

Needs https://github.com/UKHomeOffice/asl-components/pull/226

### DIFF
--- a/pages/profile/list/views/index.jsx
+++ b/pages/profile/list/views/index.jsx
@@ -65,7 +65,7 @@ function Filters() {
         formatter={filter => <Acronym>{selectivelyUppercase(filter)}</Acronym>}
         append={['pilh', 'pplh', 'admin']}
         prepend={['named']}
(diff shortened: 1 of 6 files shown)


pages/profile/list/views/index.jsx
```diff
@@ -65,7 +65,7 @@ function Filters() {
         formatter={filter => <Acronym>{selectivelyUppercase(filter)}</Acronym>}
         append={['pilh', 'pplh', 'admin']}
         prepend={['named']}
-        showAllLabel="All people"
+        showAll={{ position: 'before', label: 'Show all' }}
```
lennym, line 68: Was the intention to change the label here?

wheelsandcogs: Nope!

PR #1346, opened 2021-09-03
Don't ask newGeneticLine q unless basic or translational research

wheelsandcogs:


lennym:
I think we also need to skip the step in the setup phase.

https://github.com/UKHomeOffice/asl-pages/blob/master/pages/rops/update/config.js#L109-L112

PR #1350, opened 2021-09-07
Cat D & F notes fields are now required

wheelsandcogs:


pages/pil/procedures/schema/index.js
```diff
@@ -7,7 +7,12 @@ const procedureOptions = procedureCodes.map(code => (
     label: code + '. ' + procedureDefinitions[code],
     value: code,
     reveal: code === 'D' || code === 'F'
-      ? { notes: { inputType: 'textarea' } }
+      ? {
+        [`notesCat${code}`]: {
+          inputType: 'textarea',
+          validate: ['required']
+        }
+      }
```
lennym, line 15: Is there not some code needed somewhere to handle the fact that the field name has changed?

wheelsandcogs: I thought the same - but it does work with just this change. The revealed fields in the HTML are named `notesCatD` and `notesCatF` already.

Probably this line:
https://github.com/UKHomeOffice/asl-pages/blob/master/pages/pil/procedures/views/index.jsx?plain=1#L18
is doing the work and the "notes" key wasn't being used anywhere.

PR #1352, opened 2021-09-08
WIP sync react version and update lockfile format to v2

wheelsandcogs:


wheelsandcogs:
Will revisit when node 16 becomes LTS shortly.

PR #1397, opened 2021-11-03
Hide download links for other versions when viewing at AA est

wheelsandcogs:
Only the granted version is accessible at AA establishments, so don't provide download links for superseded or in-progress amendment versions.

### DIFF
--- a/pages/project/read/views/sections/downloads.jsx
+++ b/pages/project/read/views/sections/downloads.jsx
@@ -7,7 +7,7 @@ import { dateFormat } from '../../../../../constants';
 import { Link, Snippet, Inset } from '@asl/components';
 import Subsection from '../components/subsection';
 
-function DownloadSection({ project, version }) {
+function DownloadSection({ project, version, viewingAtAAEstablishment }) {
   const isApplication = project.status === 'inactive';
   const isAmendment = project.status === 'active' && version.status !== 'granted';
   const isGranted = version.status === 'granted';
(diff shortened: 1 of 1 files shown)


pages/project/read/views/sections/downloads.jsx
```diff
@@ -138,7 +144,8 @@ export default function Downloads() {
 
           {
             // latest version might be application, amendment or granted
-            latestVersion && <DownloadSection project={project} version={latestVersion} />
+            latestVersion &&
+              <DownloadSection project={project} version={latestVersion} viewingAtAAEstablishment={viewingAtAAEstablishment} />
```
lennym, line 148: I think for the sake of clarity I would move the check for the `viewingAtAAEstablishment` check to outside the component rather than passing in what is effectively a `dontRenderThisComponent` prop.

Otherwise it is unclear why `viewingAtAAEstablishment` is passed to some instances of `<DownloadSection />` but not others (the answer being that sometimes it should prevent rendering, but not others). Since we're already doing checks outside the component for whether it should render it is clearer to make this part of those checks.

pages/project/read/views/sections/downloads.jsx
```diff
@@ -138,7 +140,8 @@ export default function Downloads() {
 
           {
             // latest version might be application, amendment or granted
-            latestVersion && <DownloadSection project={project} version={latestVersion} />
+            latestVersion && !(grantedVersion.id !== latestVersion.id && viewingAtAAEstablishment) &&
```
lennym, line 143: What's the reasoning behind the extra `grantedVersion.id !== latestVersion.id` condition?

wheelsandcogs: If you're viewing at the primary establishment, then you should see download links to the latest version, and if the latest version is not granted, then we also need to show a section for the granted version.

If you're viewing at an AA establishment, then we should only show the granted version, regardless of amendments in progress.

PR #1407, opened 2021-11-11
Bugfix - added feature switch to enable 2022 rops

joefitter:
* ?include2022=true

### DIFF
--- a/constants/index.js
+++ b/constants/index.js
@@ -6,7 +6,6 @@ module.exports = {
     datetime: 'DD MMMM YYYY HH:mm'
   },
   ropsYears: [
-    2022,
     2021
   ]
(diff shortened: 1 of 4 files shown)


constants/index.js
```diff
@@ -6,7 +6,6 @@ module.exports = {
     datetime: 'DD MMMM YYYY HH:mm'
   },
   ropsYears: [
-    2022,
```
lennym, line 9: I think the entire PR can be replaced by some logic here that adds an extra year.

For future-proofness I'd also automatically have it include all years from 2021 until the current year, with the flag to add "next" year.

joefitter: cool - done

PR #1414, opened 2021-11-25
Feature - 2022 rops

joefitter:
* Updated rops interface to show multiple rops
* Stack previous rops in min form
* Show active rops in large form is this year or unsubmitted


joefitter:
wipping as will add unit test coverage

PR #1427, opened 2021-12-10
Show named people changes in place tasks

lennym:
When a task contained only named people changes it was super confusing.


wheelsandcogs:
Re-assigning named people on places doesn't need to be approved by ASRU - I believe there was previously a ticket to hide those changes from ASRU if there were other place amendments at the same time and the task did not auto-resolve?

What is the behaviour after this change, do the named people changes show up for ASRU?

lennym:
The changes to named people only don't need review by ASRU, but they do leave a task behind as a record, and so if they are hidden then it appears that there's a task that has no changes, and a support ticket is raised.

This change makes it so that the change is still visible so that a person looking at the resolved task can see what occurred.

PR #1428, opened 2021-12-14
clarify no procs message if nil return

joefitter:


pages/rops/procedures/list/index.js
```diff
@@ -44,6 +44,13 @@ module.exports = () => {
         res.locals.datatable.schema = pick(req.datatable.schema, ['rowNum', 'species', 'purposes', 'severity', 'severityNum']);
       }
 
+      const noProceduresCompleted = get(req, 'rop.proceduresCompleted', req.rop.proceduresCompleted) === false;
```
wheelsandcogs, line 47: is this a typo? falling back to `req.rop.proceduresCompleted` if it can't find the same?

joefitter: copy pasta from elsewhere, will fix as I'm pretty sure the fallback is pointless

PR #1432, opened 2022-01-04
Fix bugs with ROPs display for particular years

lennym:
* Don't show any year's ROP until after the 1st February of that year.
* Fix bug where 2021 ROPs were not shown for projects that ended in 2021 - calculate per year if a ROP is applicable for that year, and display a message if no years are applicable.

### DIFF
--- a/pages/project/read/views/components/rops.jsx
+++ b/pages/project/read/views/components/rops.jsx
@@ -13,7 +13,7 @@ import partition from 'lodash/partition';
 import pick from 'lodash/pick';
 import Subsection from '../components/subsection';
 
-export function Rop({ rop, project, active, ropNotRequired, url }) {
+export function Rop({ rop, project, active, url }) {
 
   const endOfYear = new Date(`${rop.year}-12-31`);
   const projEnd = project.revocationDate || project.expiryDate;
@@ -52,10 +52,6 @@ export function Rop({ rop, project, active, ropNotRequired, url }) {
         </Button>
       </form>
     );
-  }
-
-  if (ropNotRequired) {
-    return <p><strong><Snippet>rops.not-due</Snippet></strong></p>;
   }
 
   function getDeadline() {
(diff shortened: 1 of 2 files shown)


pages/project/read/views/components/rops.jsx
```diff
@@ -105,17 +101,10 @@ export function Rops({ project = {}, ropsYears = [], url, today = new Date() })
...
   const latestYear = Math.max(...ropsYears);
 
-  const deadline = subMilliseconds(new Date(`${latestYear + 1}-02-01`), 1);
-
-  const showNextYear = isAfter(today, deadline);
-  let years = ropsYears;
-
-  if (showNextYear) {
-    years = [
-      latestYear + 1,
```
lennym, line 115: This would have caused 2023 to become available on 1st Feb 2022.

pages/project/read/views/components/rops.jsx
```diff
@@ -105,17 +101,10 @@ export function Rops({ project = {}, ropsYears = [], url, today = new Date() })
...
-      ...years
-    ];
-  }
```
lennym, line 118: This code is basically all a no-op because it's always testing if today is after Feb 1st of the following year.

pages/project/read/views/components/rops.jsx
```diff
@@ -105,17 +101,10 @@ export function Rops({ project = {}, ropsYears = [], url, today = new Date() })
...
+  const years = ropsYears.filter(y => {
+    const deadline = subMilliseconds(new Date(`${y}-02-01`), 1);
+    return isAfter(today, deadline);
+  });
```
lennym, line 107: This prevents the current year from showing until Feb 1st.

pages/project/read/views/components/rops.jsx
```diff
@@ -124,20 +113,24 @@ export function Rops({ project = {}, ropsYears = [], url, today = new Date() })
 
   const rops = project.rops;
 
-  const startDate = `${thisYear}-01-01`;
   const endDate = project.revocationDate || project.expiryDate;
-  const ropNotRequired = endDate && isBefore(new Date(endDate), new Date(startDate));
```
lennym, line 129: This was the cause of the primary bug. Projects that ended prior to 2022-01-01 were not showing _any_ ROP related content.

PR #1442, opened 2022-02-11
add extra level of filters for PEL tasks

wheelsandcogs:
needs:
 - https://github.com/UKHomeOffice/asl-workflow/pull/429
 - https://github.com/UKHomeOffice/asl-workflow/pull/430

Screenshot 2022-02-14 at 12 14 02

### DIFF
--- a/pages/task/list/content/index.js
+++ b/pages/task/list/content/index.js
@@ -60,9 +60,8 @@ module.exports = {
       label: 'By PEL type:',
       options: {
         places: 'Areas',
-        roles: 'Named people',
-        amendments: 'Establishment details',
-        applications: 'Submitted applications'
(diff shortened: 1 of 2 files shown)


pages/task/list/content/index.js
```diff
@@ -55,6 +55,15 @@ module.exports = {
         ra: 'Retrospective assessments',
         changeLicenceHolder: 'Change licence holder'
       }
+    },
+    pelType: {
+      label: 'By PEL type:',
+      options: {
+        places: 'Schedule of premises',
+        roles: 'Named roles',
```
lennym, line 63: Elsewhere we refer to these as "Approved areas" and "Named people". We should to be consistent in that.

PR #1445, opened 2022-02-24
Map species codes to labels in establishment ROP download

lennym:
The labels for species as they were input are more meaningful to end users, especially in the case of the Xenopus variants where the key that is stored is actually completely incorrect (but is linked to too much data to reasonably fix forwards).


lennym:
Note that the download for individual ROPs is working as expected.

PR #1450, opened 2022-03-08
Improvements to ROPs tasks

lennym:
We want to make ROP reopening tasks visible in task histories for projects.

* Add playback of the ROPs year in task lists and task details
* Add new content for `reopened` status
* Remove logic which remaps statuses of ROPs ~because this is now done in workflow decorator~ - UPDATE: map the statuses in a server middleware here instead.
* Hide message about ROP submission date on `reopened` tasks

### DIFF
--- a/pages/task/read/views/components/task-status.jsx
+++ b/pages/task/read/views/components/task-status.jsx
@@ -3,7 +3,6 @@ import get from 'lodash/get';
 import classnames from 'classnames';
 import { Snippet } from '@asl/components';
 import Deadline from './deadline';
-import { isDeadlineExtension } from '../../../../../lib/utils';
 
 const getStatusBadge = (status, model) => {
   const good = ['resolved'];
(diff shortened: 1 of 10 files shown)


pages/task/read/views/components/task-status.jsx
```diff
@@ -14,21 +13,7 @@ const getStatusBadge = (status, model) => {
...
-  const latestActivity = task.activityLog.filter(a => a.eventName !== 'assign')[0];
-  let { action, status } = latestActivity;
-
-  if (model === 'project') {
-    const isExtension = isDeadlineExtension(latestActivity);
-
-    if (action === 'update' && isExtension) {
-      status = 'with-inspectorate';
-      action = 'with-inspectorate';
-    }
```
wheelsandcogs, line 26: Is this handled elsewhere now?

lennym: It was reading the latest activity to get the current status, but because the latest activity might not be a status change it had to do a lot of work to handle those cases.

Instead it can read the current status of the task from the `status` property and save all the work (unless there's something I've missed). Basically, the whole code block can be replaced by `const status = task.status;`

wheelsandcogs: I'm assuming at some point the task status was something other than `with-inspectorate` after an extension was triggered but if it seems fine now it probably is...

PR #1464, opened 2022-05-10
Add a unique label for role amendment resubmission

wheelsandcogs:


lennym:
This feels weirdly specific, where there's probably a case for making all of the labels for "resubmit" be similar.

PR #1491, opened 2022-07-26
Feature: refuse a project application

wheelsandcogs:
For project applications, the "rejected" option is now replaced with an "Notify of intention to refuse" which returns the application to the applicant and starts a 28 timer before the project can actually be refused. The full refusal notice is displayed in the task activity log.

Once the timer expires, the application can be refused by an inspector regardless of who the application is currently with. Refused projects will display a red banner on their landing page informing of the project status and the date when the project was refused.

### DIFF
--- a/pages/task/read/content/review.js
+++ b/pages/task/read/content/review.js
@@ -1,0 +1,14 @@
+const { merge } = require('lodash');
+const baseContent = require('./base');
+const refusalNotice = require('./refusal-notice');
+
+module.exports = merge({}, baseContent, {
+  refusalNotice,
+  page: {
+    heading: 'Check notice of intention to refuse',
+    summary: 'This notice will be sent to the applicant and a copy forwarded to the establishment licence holder and admins.'
+  },
+  buttons: {
+    submit: 'Send now'
+  }
+});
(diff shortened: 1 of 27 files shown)


pages/task/read/content/review.js
```diff
@@ -0,0 +1,14 @@
+const { merge } = require('lodash');
+const baseContent = require('./base');
+const refusalNotice = require('./refusal-notice');
+
+module.exports = merge({}, baseContent, {
+  refusalNotice,
+  page: {
+    heading: 'Check notice of intention to refuse',
+    summary: 'This notice will be sent to the applicant and a copy forwarded to the establishment licence holder and admins.'
```
lennym, line 9: I don't know if I missed something in the notifications PR, but how does this notice actually get sent to those people?

wheelsandcogs: There's a notification to external users when the task status is updated to intention-to-refuse:
https://github.com/UKHomeOffice/asl-notifications/blob/master/lib/recipients/project.js#L294-L305

and another notification to the inspector when the refusal deadline expires:
https://github.com/UKHomeOffice/asl-notifications/blob/master/lib/recipients/project.js#L144-L160

which is triggered by:
https://github.com/UKHomeOffice/asl-workflow/blob/master/bin/project-refusal-deadline-passed

PR #1494, opened 2022-08-04
Load named people only when needed

lennym:
There is an API call to `GET /establishment/:id` on every page load, which was previously loading the establishment's named people even though these were only used in a small handful of places.

In order to speed up the API calls this has been removed from the API response and split into its own endpoint which is called only on the pages where the named people are needed - specifically the establishment landing page, details page, and PDF download.

### DIFF
--- a/pages/common/middleware/index.js
+++ b/pages/common/middleware/index.js
@@ -86,22 +86,37 @@ const clearSessionIfNotFromTask = () => (req, res, next) => {
 };
 
 const populateNamedPeople = (req, res, next) => {
-  if (req.establishment) {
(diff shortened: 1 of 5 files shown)


wheelsandcogs:
Is it safe to assume you've had a check for everywhere we use `establishment.<pelh | nprc | holc | nacwo | nio | nvs | ntco | sqp>`?

lennym:
I've had a good go. Everywhere I can find that does that has the `populateNamedPeople` middleware mounted.

lennym:
Specifically: `establishment.read`, `establishment.dashboard`, `establishment.pdf`, and `task.read`

PR #1499, opened 2022-08-24
WIP Feature/suspend licence

wheelsandcogs:


pages/pil/read/views/index.jsx
```diff
@@ -184,71 +204,79 @@ export default function PIL({ pil }) {
...
-                    <Link page="task.read" taskId={openTask.id} label={<Snippet>view-task</Snippet>} className="govuk-button button-secondary" />
-                  </section>
+                  <Fragment>
+                    <section className="open-task">
+                      <h2><Snippet>{`openTask.${openTask.type}.title`}</Snippet></h2>
+                      <p><Snippet>{`openTask.${openTask.type}.description`}</Snippet></p>
+                      <Link page="task.read" taskId={openTask.id} label={<Snippet>view-task</Snippet>} className="govuk-button button-secondary" />
+                    </section>
+                    {
+                      pil.status === 'active' && isInspector && <SuspendReinstateLicence pil={pil} />
```
lennym, line 214: * There is a check here for the licence status that is not also applied for the equivalent instance of this content when there is no open task (L234)
* The display should be based on the `pil.suspend` permission rather than a specific role.

lennym: Also given the number of other bits of logic around here, it might be better to move the logic around permissions inside the `SuspendReinstateLicence` component, as it is for PPLs.

PR #1508, opened 2022-09-21
Add error message for unsupported image types

MaxOSullivan:
- Only attempt to load images which are PNG or JPEG files
- All other image types load an error message

Screenshot 2022-09-21 at 17 48 05

### DIFF
--- a/pages/project-version/docx/index.js
+++ b/pages/project-version/docx/index.js
@@ -38,6 +38,7 @@ const loadImages = attachmentHost => async node => {
   if (!url) {
     return node;
   }
+  // url is in the format data:image/type; so we use characters between index 11 and ; as the file type
   const imageType = url.substring(11, url.indexOf(';'));
   if (imageType === 'png' || imageType === 'jpeg') {
     const src = url.substring(url.indexOf(',') + 1);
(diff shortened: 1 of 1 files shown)


pages/project-version/docx/index.js
```diff
@@ -38,19 +38,27 @@ const loadImages = attachmentHost => async node => {
...
-    dimensions.width,
-    dimensions.height,
-    MAX_IMAGE_WIDTH,
-    MAX_IMAGE_HEIGHT
-  );
-  node.data.width = dimensions.width;
-  node.data.height = dimensions.height;
-  node.data.src = url;
-  return node;
+  const imageType = url.substring(11, url.indexOf(';'));
```
lennym, line 41: What's the significance of 11?

My guess is `data:image/`. If so I would probably opt for a regex match like the following to make it a bit clearer.

```suggestion
  const imageType = url.match(/^data:image\/(.+);/)[1];
```

Disclaimer: the above is totally untested.

MaxOSullivan: Have added a comment explaining, thanks for the feedback

PR #1518, opened 2022-10-13
Add assign to me link in the tasklist

wheelsandcogs:
Clicking the link assigns the selected task to the currently signed in ASRU user, but does not leave the list view.

assign-to-me

### DIFF
--- a/pages/task/list/index.js
+++ b/pages/task/list/index.js
@@ -1,9 +1,33 @@
 const { Router } = require('express');
+const bodyParser = require('body-parser');
 
-module.exports = (settings = {}) => {
+module.exports = () => {
   const app = Router();
 
   app.get('/', (req, res) => res.redirect(req.buildRoute('dashboard')));
+
+  app.put('/assign', bodyParser.json(), (req, res, next) => {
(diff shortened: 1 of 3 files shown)


pages/task/list/index.js
```diff
@@ -1,9 +1,33 @@
...
+        }
+      }
+    };
+
+    req.api(`/tasks/${taskId}/assign`, params)
+      .then(response => {
+        res.json(response.json);
+      })
+      .catch(next);
+  });
```
lennym, line 30: Is there a strong reason to add a new endpoint rather than using the existing assign endpoint?

wheelsandcogs: The current route is mounted under task.read and handles a form POST + redirect back to the task read page:
https://github.com/UKHomeOffice/asl-pages/blob/master/pages/task/read/routers/assign.js#L17-L43

We could replace the existing one with this background request one if you like?

lennym: Yeah I realised as soon as I hit send, it's a bit nasty to have two assignment routes, but not a big deal really.

PR #1522, opened 2022-10-25
Display profile merge notifications on global profiles

wheelsandcogs:


pages/global-profile/views/index.jsx
```diff
@@ -33,6 +34,29 @@ const selector = ({
 
 const formatDate = (date, format) => (date ? dateFormatter(date, format) : '-');
 
+const ProfileMerges = ({ profile }) => {
+  if (profile.profileMerges.length === 0) {
```
lennym, line 38: Just in case it's not there at all.

```suggestion
  if (!profile.profileMerges || profile.profileMerges.length === 0) {
```

PR #1523, opened 2022-10-26
Add block class to fix formatting in larger table

MaxOSullivan:
Give 'unassigned' block formatting so it stays on its own line no matter how big the table cell is

Screenshot 2022-10-26 at 11 28 15

### DIFF
--- a/pages/task/list/components/assign-task.jsx
+++ b/pages/task/list/components/assign-task.jsx
@@ -44,7 +44,7 @@ export default function AssignTask({ task }) {
 
   return (
     <Fragment>
-      <em>Unassigned</em>
+      <em className="block">Unassigned</em>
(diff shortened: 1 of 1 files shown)


pages/task/list/components/assign-task.jsx
```diff
@@ -44,7 +44,7 @@ export default function AssignTask({ task }) {
 
   return (
     <Fragment>
-      <em>Unassigned</em>
+      <em className='block'>Unassigned</em>
```
lennym, line 47: I'm a little surprised that passed linting.
```suggestion
      <em className="block">Unassigned</em>
```

MaxOSullivan: Sorted

PR #1525, opened 2022-10-28
Better PIL amend description for ASRU users

wheelsandcogs:
The previous description stated that they could not transfer the licence, when they actually can do.

### DIFF
--- a/pages/pil/read/index.js
+++ b/pages/pil/read/index.js
@@ -65,7 +65,6 @@ module.exports = settings => {
     res.locals.static.profile = req.profile;
     res.locals.static.currentPath = req.originalUrl;
     res.locals.static.isLicenceHolder = req.user.profile.id === req.profileId;
-    res.locals.static.isAsruUser = !!req.user.profile.asruUser;
     next();
   });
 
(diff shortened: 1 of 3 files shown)


pages/pil/read/index.js
```diff
@@ -65,6 +65,7 @@ module.exports = settings => {
     res.locals.static.profile = req.profile;
     res.locals.static.currentPath = req.originalUrl;
     res.locals.static.isLicenceHolder = req.user.profile.id === req.profileId;
+    res.locals.static.isAsruUser = !!req.user.profile.asruUser;
```
lennym, line 68: This data point is already available on the client at `static.profile.asruUser` so it's better to use that rather than add _yet another data point_.

wheelsandcogs: Yeah `req.profile` in this case is the PIL holder, not the currently logged in user.

I thought it was as well, but both `isAsru` and `isAsruUser` were undefined until I added this.

lennym: See my eventual edit above.

PR #1533, opened 2022-11-14
fix different content warnings between server / client

wheelsandcogs:


lennym:
How does this make any difference to anything? Looks like a no-op to me.

wheelsandcogs:
React doesn't like the `<thing>Plaintext {var}</thing>` format.

Before:
Screenshot 2022-11-15 at 09 25 13

After:
Screenshot 2022-11-15 at 09 26 28

lennym:
Weird. That's basically [the canonical way to write React](https://reactjs.org/tutorial/tutorial.html#what-is-react).

wheelsandcogs:
Maybe it's something peculiar with ROPs props and this is just a side effect? Happy to close and ignore it, just saw the warning as I was going through testing.

lennym:
I wonder if it's to do with the fact that both of those values will be numbers, and so interpolating them into a string literal makes it go away. I can't believe this is the _only_ time we're writing that kind of code.

wheelsandcogs:
Sounds plausible

PR #1535, opened 2022-12-07
Add conditions during PEL amendments

MaxOSullivan:
- Add conditions to PEL task pages
- Edit task router to include condition and reminder data

### DIFF
--- a/pages/task/read/views/models/establishment.jsx
+++ b/pages/task/read/views/models/establishment.jsx
@@ -4,17 +4,21 @@ import {
   StickyNavAnchor,
   Snippet,
   Diff,
-  DiffText
+  DiffText,
+  Conditions
 } from '@asl/components';
 import { hasChanged } from '../../../../../lib/utils';
 import establishmentSchema from '../../../../establishment/update/schema';
 import NamedPeople from '../../../../establishment/components/named-people';
 import formatters from '../../../../establishment/formatters';
 import Authorisations from '../../../../establishment/update/views/authorisations';
+import isEmpty from 'lodash/isEmpty';
 
 export default function Establishment({ task, values }) {
-  const establishment = useSelector(state => state.static.establishment);
+  const { establishment, allowedActions, openTask, errors } = useSelector(state => state.static);
   const isComplete = !task.isOpen;
+  const canUpdateConditions = allowedActions.includes('establishment.updateConditions');
+  const taskData = task.data.data;
 
   return [
     (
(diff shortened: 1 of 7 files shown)


pages/task/read/views/models/establishment.jsx
```diff
@@ -52,6 +56,21 @@ export default function Establishment({ task, values }) {
             proposedLabel={isComplete ? <Snippet>diff.changed-to</Snippet> : undefined}
           />
 
+          <div className="sticky-nav-anchor">
+            <h2><Snippet>conditions.title</Snippet></h2>
+            <Conditions
+              conditions={taskData.conditions ? taskData.conditions : establishment.conditions }
+              reminders={taskData.reminder !== '' ? [JSON.parse(taskData.reminder)] : establishment.reminders}
```
gaz1634, line 63: Does this work if there are no reminders in the task?
Is the JSON.parse necessary?

pages/task/read/views/models/place.jsx
```diff
@@ -97,6 +100,22 @@ export default function Playback({ task, values, allowSubmit }) {
           setDirty={setDirty}
         />
       </StickyNavAnchor>
+    ),
+    (
+      <StickyNavAnchor id="conditions" key="conditions">
+        <h2><Snippet>conditions.title</Snippet></h2>
+        <Conditions
+          conditions={taskData.conditions ? taskData.conditions : establishment.conditions }
+          reminders={taskData.conditions ? [JSON.parse(taskData.reminder)] : establishment.reminders}
```
gaz1634, line 109: Similar to my previous comment. Is parse necessary?

Also should the logic on populating the Condition components reminders property be consistent?

pages/task/read/views/models/role.jsx
```diff
@@ -96,6 +99,22 @@ export default function Role({ task, values, schema }) {
           }
         </StickyNavAnchor>
       )
+    ),
+    (
+      <StickyNavAnchor id="conditions" key="conditions">
+        <h2><Snippet>conditions.title</Snippet></h2>
+        <Conditions
+          conditions={taskData.conditions ? taskData.conditions : establishment.conditions }
+          reminders={taskData.conditions ? [JSON.parse(taskData.reminder)] : establishment.reminders}
```
gaz1634, line 108: Again, should this logic be consistent with other reminders properties?

pages/task/read/views/models/establishment.jsx
MaxOSullivan: Have made the passing of reminders consistent
Parse is needed due to the way the data is passed through at the bottom of this file https://github.com/UKHomeOffice/asl-components/pull/277/files#diff-702fbff05fa84184acb68e5f199300abac056b3eabe558361ded0327cafce9f3

The object needs to be added as part of the form to be sent in the request when the amendment is accepted or returned
Happy to change it if there's an easier way

gaz1634 (approved):

PR #1554, opened 2023-06-20
BUGFIX ASL-4331 check if "other establishments" is set to false

roc:
STUB

Add description and screenshots


easternbloc (approved):
LGTM

joefitter (approved):
:-)

PR #1567, opened 2023-11-06
Revert "Bugfix/assb 1345 without diff window regression"

taomoh:
Reverts https://github.com/UKHomeOffice/asl-pages/pull/1566

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@asl/pages",
-  "version": "30.2.4",
+  "version": "30.2.8",
   "description": "",
   "main": "index.js",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@asl/pages",
-  "version": "30.2.7",
+  "version": "30.2.4",
```
jeff-horton-ho-sas, line 3: Will this work with newer versions already published to npm - should we instead be releasing a newer version with the reverted code?

jeff-horton-ho-sas (approved):

PR #1593, opened 2024-05-30
[ASL-4537] Text change to current billing period

jeff-horton-ho-sas:


edemirbag (approved):

pages/establishment/licence-fees/details/read/views/index.jsx
```diff
@@ -19,9 +19,6 @@ export default function Details() {
     declaredCurrent: {
       format: (value, model) => {
         return model.declaredCurrent ? 'Yes' : 'No';
-      },
```
farrukhishaq, line 22: How does this change meets AC02. I may have missed understanding this ticket.

jeff-horton-ho-sas: This was written before AC02 was added. I think the AC is wrong, I'll update if needed when I get an answer.

PR #1610, opened 2024-08-25
Display Removed flag in protocols' steps

parvinba:


pages/project-version/middleware/index.js
```diff
@@ -304,7 +304,8 @@ const getPreviousProtocols = (firstVersion, previousVersion, grantedVersion) =>
   const previous = get(previousVersion, 'data.protocols', []).filter(Boolean).filter(p => !p.deleted).map(p => p.id);
   const granted = get(grantedVersion, 'data.protocols', []).filter(Boolean).map(p => p.id);
   const showDeleted = uniq([ ...previous, ...granted ]);
-  return { first, previous, granted, showDeleted };
+  const steps = get(previousVersion, 'data.protocols', []).filter(Boolean).filter(p => !p.deleted).map(p => p.steps);
```
farrukhishaq, line 307: Would it be helpful to use the element's name to determine the actual key instead of `p`?

package.json
```diff
@@ -26,7 +26,7 @@
   },
   "homepage": "https://github.com/UKHomeOffice/asl-pages#readme",
   "dependencies": {
-    "@asl/projects": "^15.5.15",
+    "@asl/projects": "^15.5.16",
```
jeff-horton-ho-sas, line 29: This will need to be 15.5.17 when asl-projects has been merged and tagged.

Also make sure to run `npm install` and commit the updated package-lock.json when updating dependencies, or `npm ci` won't work in the pipeline.

parvinba: updated to latest

KiranVarma02 (approved):

PR #1620, opened 2024-09-16
dont merge

parvinba:


pages/project-version/middleware/index.js
mohomeoffice, line 1: Parvin could you please make sure we do not have any redundant checks, as it would make execution slower, also there is no default handler - I am approving it but for the future please

mohomeoffice (approved):
Parvin could you please make sure we do not have any redundant checks, as it would make execution slower, also there is no default handler - I am approving it but for the future please

PR #1621, opened 2024-09-17
router to use route path.

farrukhishaq:
path to regex issue, unknow. 

https://github.com/expressjs/express/issues/5955


mohomeoffice (approved):
lets see if it works

PR #1633, opened 2024-10-03
[ASSB-1418] Upstream fixes accessing fields of undefined when version is RA

jeff-horton-ho-sas:


mohomeoffice (approved):
looks okay to me

PR #1641, opened 2024-10-24
swapping true and false endangered display - rop - add procedure

parvinba:


mohomeoffice (commented):
},
        false 
        
        instead of } 
        
        Could you please check again

parvinba:
> }, false
> 
> ```
>     instead of } 
>     
>     Could you please check again 
> ```

Last element in array don't need a comma. This was case before swell, I just swapped the order.

mohomeoffice (approved):
looks okay to me

PR #1642, opened 2024-10-30
Updating ROP endangered feature ASL-4448

parvinba:


mohomeoffice (commented):
// clear endangered details if user selects 'No' for endangered
if (!values.endangered) {
  values.endangeredDetails &&= null;
}
Could you please try it like this

parvinba:
if (!values.endangered) was done. second line is not clear, what u wanted, left it as it is, as it is working as expected,

mohomeoffice (approved):
looks okay to me

PR #1658, opened 2024-11-28
asl-4594 AC04 updated to match confluence

parvinba:


pages/training/modules/views/index.jsx
```diff
@@ -11,6 +11,7 @@ export default function Type() {
         title={<Snippet isExemption={model.isExemption}>title</Snippet>}
         subtitle={`${profile.firstName} ${profile.lastName}`}
       />
+      <p><Snippet>hint</Snippet></p>
```
kiran-varma-home-office, line 14: shouldn't it be `{hint}` ?

parvinba: @kiran-varma-home-office, just hint will only work

jeff-horton-ho-sas: @kiran-varma-home-office Snippet takes it's children as a string and looks up that snippet. so `<Snippet>hint</Snippet>` is correct.

jeff-horton-ho-sas (approved):

PR #1666, opened 2024-12-11
asl-4629-endangered-cited

parvinba:


pages/rops/update/schema/index.js
```diff
@@ -19,6 +19,14 @@ module.exports = req => {
     return opt;
   };
 
+  const disableProcRadio = fieldName => {
+    const field = req.rop.procedures.map(p => p[fieldName]);
+    if (field.includes(true)) {
+      return true;
+    }
+    return false;
```
jeff-horton-ho-sas, line 27:
```suggestion
    return field.includes(true);
```

pages/rops/update/schema/index.js
```diff
@@ -140,7 +148,11 @@ module.exports = req => {
       format: toBoolean,
       automapReveals: true,
       options: [
-        false,
+        {
+          value: false,
+          disabled: disableProcRadio('endangered'),
+          warning: content.fields.disabledWarning
```
jeff-horton-ho-sas, line 154: Does the warning need to be conditional on the field being disabled?

pages/rops/update/schema/index.js
parvinba: good suggestion.. done

pages/rops/update/schema/index.js
parvinba: good suggestion.. done

jeff-horton-ho-sas: https://github.com/UKHomeOffice/react-components/blob/main/src/forms/radio-group.jsx#L72-L75

This should either be renamed to `disabledWarning` or it should always display the warning an the warning is only passed if needed, i.e.
```js
{
    opt.warning &&
    <div className="govuk-reveal">
        <Warning>{opt.warning}</Warning>
    </div>
}
```
I think I prefer the second option as it makes it possible to have warnings that always show.

pages/rops/update/schema/index.js
```diff
@@ -140,7 +145,11 @@ module.exports = req => {
       format: toBoolean,
       automapReveals: true,
       options: [
-        false,
+        {
+          value: false,
+          disabled: disableProcRadio('endangered'),
+          ...(disableProcRadio('endangered') && {warning: content.fields.disabledWarning})
```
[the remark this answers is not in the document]
jeff-horton-ho-sas: Further thoughts this could then be 
```suggestion
          reveal: disableProcRadio('endangered') ? <Warning><Snippet>fields.disabledWarning</Snippet></Warning> : undefined
```
And we don't need the very specific addition of a warning reveal to the generic react-component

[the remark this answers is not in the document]
parvinba: Yes I am raising another PR for react-component to make it generic.. as suggested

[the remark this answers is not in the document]
jeff-horton-ho-sas: I'd say just undo the previous changes to react-components and use the existing reveal prop?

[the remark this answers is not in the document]
parvinba: https://github.com/UKHomeOffice/react-components/pull/102

mohomeoffice (approved):
looks okay to me

PR #1669, opened 2025-01-16
feat: [ASL-4594] Named person guides MVP1 - link in person screen

kiran-varma-home-office:
Screenshot 2025-01-16 at 11 14 09


jeff-horton-ho-sas (commented):
Please see comment on rel attribute, otherwise this is good to go :)

jeff-horton-ho-sas (approved):

PR #1678, opened 2025-02-25
Bug/asl 4528 correct hyper link

farrukhishaq:
made the use of model data and filtering on transferedProject to update and correctly assign the projected when the transfer journey is resolved.


mohomeoffice (changes requested):
Build failed

PR #1688, opened 2025-03-06
Keycloak V22

farrukhishaq:
Keycloak Realm now using lower case, Keycloak URL now using V22 instance URI, mocha also updated.


jeff-horton-ho-sas (approved):

mohomeoffice (approved):
looks okay to me

